PR #2, opened 2019-04-14
More advanced authentication

### LINKED ISSUE #1: [RFC] Use Role Based Security
Inspired by [HashiCorp Vault policies](https://www.vaultproject.io/docs/concepts/policies.html#policy-syntax).

This allows to finely manage API route access, using patterns and permission composition.
The default permission is to deny the access to the resource.

This implementation isn't optimized, we can cache the roles in to the user struct in the future.

In the future we can even allow/disallow parameters like `?host=` or `?port=`

```cr
require "json"

struct Document
  include JSON::Serializable
  getter groups : Array(Group)
  getter users : Array(User)

  struct User
    include JSON::Serializable
    getter name : String,
      id : String,
      group_ids : Set(String)
  end

  struct Group
    include JSON::Serializable
    getter name : String,
      id : String,
      description : String,
      path_policies : Hash(String, PathPolicy)

    def path_permissions(path : String, &block : Set(PathPolicy::Permission) ->)
      path_policies.each do |path_key, policy|
        if File.match? path_key, path
          yield policy.permissions
        end
      end
    end

    struct PathPolicy
      include JSON::Serializable
      getter permissions : Set(Permission)

      enum Permission
        Create
        Read
        Update
        Delete

        def self.new(permission : String)
          {% begin %}
          case permission
            {% for const in @type.constants %}
          when {{const.downcase}} then {{const}}
            {% end %}
          else raise "invalid permission: #{permission}"
          end
          {% end %}
        end

        def to_json(json : JSON::Builder)
          {% begin %}
          case self
            {% for const in @type.constants %}
          when {{const}} then json.string "{{const.downcase}}"
            {% end %}
          else raise "unkown permission: #{value}"
          end
          {% end %}
        end
      end
    end
  end

  def path_permissions(user : String, path : String) : Set(Group::PathPolicy::Permission)?
    puts path
    permissions = nil

    if user = @users.find &.name.== "Tom"
      @groups.each do |group|
        if user.group_ids.includes? group.id
          group.path_permissions path do |group_permissions|
            if permissions
              permissions = permissions & group_permissions
            else
              permissions = group_permissions
            end
          end
        end
      end
    end
    if !permissions || permissions.try &.empty?
      puts "access to `#{path}` denied to #{user}"
      return
    end
    puts permissions
  end
end

json = <<-JSON
{ "users" : [
    {
       "name": "Tom",
       "id": "092",
       "group_ids": ["123"]
    }
  ],
  "groups": [
     {
      "name": "User",
      "id": "123",
      "description": "Default user group",
      "path_policies": {
        "/**": {
          "permissions": ["create", "read", "update", "delete"]
        },
        "/app/**": {
          "permissions": ["read", "update"]
        },
        "/app/*-private": {
          "permissions": []
        },
        "/app/dppm": {
          "permissions": []
        }
      }
    }
  ]
}
JSON

doc = Document.from_json json

doc.path_permissions "Tom", "/app"
doc.path_permissions "Tom", "/app/dppm"
doc.path_permissions "Tom", "/app/new-app"
doc.path_permissions "Tom", "/app/new-private"

```
Output:
``` 
/app
Set{Create, Read, Update, Delete}
/app/dppm
access to `/app/dppm` denied to Document::User(@name="Tom", @id="092", @group_ids=Set{"123"})
/app/new-app
Set{Read, Update}
/app/new-private
access to `/app/new-private` denied to Document::User(@name="Tom", @id="092", @group_ids=Set{"123"})
```

dscottboggs:
This is still somewhat of a WIP -- it does not yet compile due to some confusion with macros that I'm too tired to figure out at the moment.  However, the format has basically been figured out internally, for the config file, and for the REST API, so merging is an imminent possibility. Comments and suggestions welcome.

This will resolve #1.


api-options.yml
```diff
@@ -0,0 +1,108 @@
...
+            example:
+              - path pattern: /*/some/{other,options}/**
+                permissions: Create | Read | Update | Delete
+              - path pattern: /another/path
+                permissions: None
+
+          GET:
+            returns: The group's configuration
+
+      /user:
```
j8r: Why this file is there?

src/dppm_rest_api.cr
```diff
@@ -2,8 +2,9 @@ require "./utils"
 require "./dppm_rest_api/actions"
 
 module DppmRestApi
-  VERSION  = "0.1.0"
+  VERSION  = "0.2.0"
```
j8r: That's not needed. Gemfile habit?

src/dppm_rest_api.cr
```diff
@@ -2,8 +2,9 @@ require "./utils"
...
   TEST_DIR = "#{File.dirname(__DIR__)}/test-data"
```
j8r: `test-data`... test related stuff should go in `spec`, that's where this data is used.

src/config/user.cr
```diff
@@ -0,0 +1,81 @@
...
+         (name = data["name"]?).is_a?(String) &&
+         (key = data["API key hash"]?).is_a? String
+        new key, groups.split(',').map(&.to_i), name
+      end
+    rescue ArgumentError
+      nil
+    end
+
+    private def serialized_groups : String
+      groups.map(&.to_s base: 16).join(",")
```
j8r: can be a comma :)

src/config/route.cr
```diff
@@ -0,0 +1,58 @@
...
+
+module DppmRestApi
+  # Matches a request to a series of globs loaded from a config file.
+  struct Route
+    # The permission/access level for this route
+    property access : Access
+
+    property query_parameters : Hash(String, Array(String))
+
+    macro finished
```
j8r: What did you try to achieve? Including `JSON::Serializable`doesn't work>

j8r (changes requested):

j8r:
This PR is impressive! :o

api-options.yml
dscottboggs: I thought it should be kept track of as a part of version control. Do you disagree?

src/config/user.cr
dscottboggs: What do you mean?

api-options.yml
j8r: This file isn't documented, and is a draft that will disappear in he future. Furthermore, this will conflict more or less with the autogenerated documentation in the future.
If you really want to add it somewhere, a gist is the best place IMHO (because of comments), the wiki second.

src/config/user.cr
j8r: Sorry, I mean the comma can be a char

j8r:
This branch needs to be rebased.

dppm-api.html
```diff
@@ -138,6 +138,6 @@
...
...&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; returns: list of package IDs<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; DELETE:<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; does: clean unused built packages<br>&nbsp;&nbsp;&nbsp; &#x2F;:id&#x2F;:version:<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; required: id<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; optional: version<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; verbs:<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; GET:<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; returns: JSON-formatted output of `dppm package query :id`<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; DELETE:<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; does: removes the given package<br>&nbsp;&nbsp;&nbsp; &#x2F;build&#x2F;:package_name&#x2F;:version:<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; required: package_name<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; optional: version<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; verbs: POST<br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; returns: the ID of the built image<br>&nbsp;&nbsp;&nbsp;&nbsp;<br>&nbsp; &#x2F;src?type={package_type}:<br>&nbsp;&nbsp;&nbsp; # list all available source packages<br>&nbsp;&nbsp;&nbsp; # the package type is either &quot;app&quot; or &quot;lib&quot;<br><br>
```
j8r: what is it?

src/config.cr
```diff
@@ -1,2 +1,126 @@
...
+    end
+
+    new (file_loc || Defaults.file_loc).not_nil!,
+      (host || Defaults.host).not_nil!,
+      (port || Defaults.port).not_nil!,
+      (public_host || Defaults.public_host), # may be nil
+      (default_namespace || Defaults.default_namespace).not_nil!,
+      (log_file || Defaults.log_file).not_nil!
+  end
+end
```
j8r: This file isn't needed, the config parsing is done by the DPPM library along with the CLI https://github.com/DFabric/dppm-rest-api/blob/master/src/cli.cr

api-options.yml
j8r: Still not removed. We'll have a cloud for this type of documents.

j8r (changes requested):

dscottboggs:
There are a number of additional options supported by the config on this branch than the master branch has. Perhaps we could merge this into master, then work on converting the configuration to work together with DPPM proper as a separate issue? The branch runs and passes specs in its current state. 

See https://github.com/DFabric/dppm-rest-api/blob/d9a13cc8079e4ab834418531da233acfafd4e294/src/config.cr#L42-L54 and https://github.com/DFabric/dppm-rest-api/blob/master/src/cli.cr

j8r:
There are no problems on my side, but there are several issues before being technically being able to merge:
- Rebase against master
- CI green

j8r:
Personally there are no problems on my side, but there are several issues before technically being able to merge:
- Rebase against master and resolve conflicts
- CI green
Then we are ready to merge.

j8r:
You shouldn't have merged this @dscottboggs , this totally fucked up the master history.

Issue #4, opened 2019-04-28
Update configuration to use Clicr; integrate the repository with DPPM proper


dscottboggs:
If you wouldn't mind, I could use a hand setting up the config file/CLI/internal config structure to match up with the general style of DPPM and integrate with it. See my update to `cli.cr` in [a402837d](https://github.com/DFabric/dppm-rest-api/commit/a402837d25e3de179b86a97180a3a3d9e7ef0cef).

dscottboggs:
resolved by #8

PR #5, opened 2019-04-28
More advanced auth

dscottboggs:


j8r:
Reborn of https://github.com/DFabric/dppm-rest-api/pull/2

src/dppm_rest_api/actions/app.cr
```diff
@@ -1,4 +1,7 @@
 require "../../utils"
+require "dppm"
+require "../../config/route"
+require "../../utils"
```
j8r: You shouldn't backward require modules, that the purpose of organizing our code in trees.
This simplify testing and refactoring, especially when changing paths

config.con
```diff
@@ -1,5 +1,3 @@
 host "[::1]"
 port 8994
-sources {
-  default "https://github.com/DFabric/packages-source/tarball/master"
-}
+source "https://github.com/DFabric/packages-source/tarball/master"
```
j8r: the last one was correct, this change is in DPPM@master

dscottboggs: ok thanks I was wondering which of those it was

dscottboggs: I changed it back to `sources {...}` and even after running `shards update` I got an exception -- `Unhandled exception: Unexpected token: EndHash (Exception)`

j8r: Use this one https://github.com/DFabric/dppm/blob/master/config.con.
I also made a new commit to bump to Crystal 0.28.0, which is now in Alpine

dscottboggs: Thank you!

dscottboggs:
CI is still failing due to #4.

j8r:
No @dscottboggs , that's because the CI is now on Crystal 0.28.0. I've pushed a commit to fix the URI changes.

dscottboggs:
I see. I have updated, but the CI is still failing due to a missing file. I'm not sure what that file is or why it's causing an exception.

j8r (approved):

PR #6, opened 2019-05-05
Reorganize project

j8r:
Move files to the top level, and use more features of the DPPM library like the CLI


src/cli.cr
```diff
@@ -11,14 +11,20 @@ module DppmRestApi::CLI
       port ||= dppm_config.port
       host ||= dppm_config.host
     end
-    DppmRestApi.run port, host
+    DppmRestApi.run host, port, data_dir
```
dscottboggs, line 14: The other options are already set by DPPM?

src/route.cr
```diff
@@ -0,0 +1,28 @@
+require "./access"
```
dscottboggs: I think this file should be namespaced under `DppmRestApi::Config`, and be at src/config/route.cr.

dscottboggs (changes requested):
Thanks for bringing the inconsistency to my attention :+1:

src/cli.cr
j8r: Yes, this variables comes from the CLI, which has default values (below).

j8r:
Changes done, as you suggested (I killed the first CI job, the other one pass).

dscottboggs (approved):

PR #7, opened 2019-05-10
Add linter

dscottboggs:


dscottboggs:
OH COME ON

src/actions/app.cr
```diff
@@ -183,16 +174,17 @@ module DppmRestApi::Actions::App
   end
   # Stream the logs for the given application over the websocket connection.
   relative_ws "/:app_name/logs" do |sock, context|
-    app_name = context.params.url["app_name"]
     if context.current_user? && Config.has_access? context, Access::Read
       # TODO: stream logs to sock
+      sock.tap do
+        # trick ameba into ignoring the sock variable
+      end
```
j8r: ... does renaming `sock` into `_`  can work, too?

j8r (commented):
You can also `shards update`, I've pushed new commits into dppm@master.

src/actions/app.cr
dscottboggs: it does. The reason I didn't do that was I wanted a note of what it was for later, but I can just as easily add a comment.

j8r: Having a comment sounds better than having a dummy code specifically for ameba.

dscottboggs: I was actually kind of hesitant to add Ameba to the CI because of situations like this, but I discovered a few bugs in the process of adding it so it seems worth it.

j8r (approved):

j8r:
Nice work and commits @dscottboggs , thanks!

PR #8, opened 2019-05-11
Integrate CLI with DPPM proper; add user/group modification commands

dscottboggs:


src/cli.cr
```diff
@@ -40,6 +167,179 @@ CLI.run(
...
+            member of. It is not recommended to match only by name or by groups,
+            as your command may unintentionally affect similar users. For
+            example, two users may choose the same name, but be members of
+            different groups. This doesn't matter as they're only identified
+            internally by their API keys, but you may unintentionally add/remove
+            groups from those users when using this command in that situation.
+
+            Ideally, you won't have multiple users with the same name, or you
+            can use the web interface to edit these values using API keys for
+            exact identification.
```
j8r: Absolutely too long, users won't want to have their terminals taken with textual information.
This documentation has its place in a `man` or a proper documentation place (web, markdown...)

src/cli.cr
```diff
@@ -40,6 +167,179 @@ CLI.run(
...
+              },
+              match_groups: {
+                info: <<-HERE,
+                The groups selected users must be a member of. They may
+                also be members of grousp that are not specified.
+                You may specify several groups by separating them with
+                commas, and multiple group selectors by separating
+                them with a colon, like `match-groups=1,2,3:2,4', which
+                would select a user who's a member of groups 1, 2, and
+                3, or a user who's a member of groups 2 and 4.
```
j8r: Same here, way too long

src/cli.cr
```diff
@@ -40,6 +167,179 @@ CLI.run(
...
+            member of. It is not recommended to match only by name or by groups,
+            as your command may unintentionally affect similar users. For
+            example, two users may choose the same name, but be members of
+            different groups. This doesn't matter as they're only identified
+            internally by their API keys, but you may unintentionally add/remove
+            groups from those users when using this command in that situation.
+
+            Ideally, you won't have multiple users with the same name, or you
+            can use the web interface to edit these values using API keys for
+            exact identification.
```
j8r: Way too long. Look at other clis in your system as examples.

j8r (changes requested):
The cli information should be a short description of the command, option or variable. Else, the terminal is filled with too much text.
If the user doesn't know what a command do, (s)he can refer to the docs where details are explained.

src/cli.cr
```diff
@@ -40,6 +231,235 @@ CLI.run(
...
+                       They may also be members of groups that are not \
+                       specified. You may specify several groups by separating \
+                       them with commas, and multiple group selectors by \
+                       separating them with a colon, like \
+                       `match-groups=1,2,3:2,4', which would select a user \
+                       who's a member of groups 1, 2, and 3, or a user who's a \
+                       member of groups 2 and 4."
+              },
+              api_key: {
+                short: "key",
```
j8r: Why having this? `short` is only useful for options, usually a char.

src/cli.cr
dscottboggs: Updated as requested

src/cli.cr
```diff
@@ -40,6 +256,180 @@ CLI.run(
...
+            variables: {
+              group_id: {
+                info: "the numeric ID of the group to edit",
+              },
+              path: {
+                info: "the path on which to update allowed query parameters.",
+              },
+              key:      {info: "the query key to match for"},
+              add_glob: {
+                short: "add",
```
j8r: Useless

src/cli.cr
```diff
@@ -40,6 +256,180 @@ CLI.run(
...
+            action:    "DppmRestApi::CLI.edit_users",
+            variables: {
+              match_name: {
+                info: "the name of the users to filter by.",
+              },
+              match_groups: {
+                info: "The groups that selected users must be a member of.",
+              },
+              api_key: {
+                short: "key",
```
j8r: Useless

src/cli.cr
```diff
@@ -40,6 +256,180 @@ CLI.run(
...
+                info:  "the api key to select a user by.",
+              },
+              output_file: {
+                short: "file",
```
j8r: Ditto

src/cli.cr
```diff
@@ -40,6 +256,180 @@ CLI.run(
...
+            alias:     "rm",
+            variables: {
+              match_name: {
+                info: "the name of the users to filter by.",
+              },
+              match_groups: {
+                info: "The groups that selected users must be a member of.",
+              },
+              api_key: {
+                short: "key",
```
j8r: Ditto

src/cli.cr
```diff
@@ -40,6 +256,180 @@ CLI.run(
...
+            action:    "DppmRestApi::CLI.show_users",
+            variables: {
+              match_name: {
+                info: "the name of the users to filter by.",
+              },
+              match_groups: {
+                info: "The groups that selected users must be a member of.",
+              },
+              api_key: {
+                short: "key",
```
j8r: Ditto

j8r (dismissed):
The variables are applied recursively, you don't need to copy yourself.

j8r:
Trailing dots can also be removed, they aren't commonly present for short descriptions.

j8r:
`Clicr` updated to `0.6.0`, you can now use `description` with heredocs for long descriptions.
As you wish if you prefer to keep this information in the README of to put them in the cli.

j8r:
Resolves https://github.com/DFabric/dppm-rest-api/issues/4

src/cli.cr
```diff
@@ -40,6 +270,149 @@ CLI.run(
...
+            alias:     "q",
+            action:    "DppmRestApi::CLI.edit_group_query",
+            variables: {
+              group_id: {
+                info: "the numeric ID of the group to edit",
+              },
+              path: {
+                info: "the path on which to update allowed query parameters.",
+              },
+              key:      {info: "the query key to match for"},
```
j8r: put it in a multi line

src/cli.cr
```diff
@@ -40,6 +270,149 @@ CLI.run(
...
+              permissions: {
+                info: "\
+                  JSON-formatted permissions, mapping a path to the query \
+                  parameters and type of access allowed on that path",
+                default: {
+                  "/**"                       => DppmRestApi::Config::Route.new(DppmRestApi::Access.deny),
+                  "/{app,pkg,src,service}/**" => DppmRestApi::Config::Route.new(
+                    DppmRestApi::Access::All,
+                    {"namespace" => ["default-namespace"]}),
+                }.to_json,
```
j8r: That's weird. Can't it be a constant somewhere, I suppose it can be reusable it other places?

src/cli.cr
```diff
@@ -40,6 +270,149 @@ CLI.run(
           },
         },
       },
+      add_user: {
+        info:      "add a user",
+        action:    "DppmRestApi::CLI.add_user",
```
j8r: `DppmRestApi::CLI.` can be dropped (see also all other places below)

j8r (commented):
What is the `mn_`? Can't it be just `data_dir` and `groups`?
Or, you can have `raw_` appended to the variable to underline the fact it's a raw string.

src/cli.cr
```diff
@@ -2,7 +2,80 @@ require "dppm/cli"
...
+  private macro required(arg)
+    mn_{{arg.id}} || abort "{{ arg.id.gsub /_/, "-" }} argument is required."
+  end
+
+  # select the users from the current configuration according to the name,
+  # groups the user is a member of, and/or the user's API key.
+  private def select_users(match_name, match_groups, api_key, from selected_users)
+    # Filter by name
+    if name = match_name
+      if name.starts_with?('/') && name.ends_with?('/')
```
j8r:
```suggestion
      if stripped_name = name.lchop?('/').try &.rchop?('/')
```

src/cli.cr
```diff
@@ -2,7 +2,80 @@ require "dppm/cli"
...
+        # match by regex
+        name_regex = Regex.new name.lchop('/').rchop('/')
```
j8r:
```suggestion
        name_regex = Regex.new stripped_name
```

src/cli.cr
```diff
@@ -2,7 +2,80 @@ require "dppm/cli"
...
+          end
+        end
+      end
+    end
+    # filter by API key if specified.
+    if key = api_key
+      selected_users = selected_users.select { |user| user.api_key_hash == key }
+      # freak out if multiple users were found.
+      if selected_users.size > 1
+        abort String.build { |msg|
```
j8r: Use `do ... end`

src/cli.cr
dscottboggs: Fair point

src/cli.cr
dscottboggs: *Very* nice. I had a feeling I didn't have the best solution.

src/cli.cr
dscottboggs: I used curly-braces in this case because it has different precedence than `do..end` I will save as an intermediate variable to avoid confusion

src/cli.cr
dscottboggs: It can just be `action: :add_user`?

src/cli.cr
```diff
@@ -13,6 +86,163 @@ module DppmRestApi::CLI
...
+    current_config = File.open permissions_file do |file|
+      DppmRestApi::Config.from_json file
+    end
+    JSON.build STDOUT, indent: 2 do |builder|
+      current_config.users.to_json builder
+    end
+    puts
+  end
+
+  def add_group(id mn_id, name mn_name, permissions mn_permissions, data_dir mn_data_dir, **args)
```
j8r: It would be simpler to have `add_group(id, name, permissions, data_dir, **args)`, the `String | Nil` -> `String` cast doesn't need an intermediate variable.

dscottboggs: I don't get what you mean? Of course a cast from nil requires an intermediate variable?

src/cli.cr
dscottboggs: I tried setting the values like this and it said `undefined method 'run_server' for CLI:Module`

src/cli.cr
j8r: No:
```cr
name = if true
  ""
else
  nil
end
puts typeof(name)
name || raise "that's nil!"
puts typeof(name) #=> String
```

j8r: No:
```cr
name = if true
  ""
else
  nil
end
puts typeof(name) #=> (String | Nil)
name || raise "that's nil!"
puts typeof(name) #=> String
```

src/cli.cr
j8r: That depends where the macro is expanded.
I think it can be inside the `CLI` module, inside `def run`, then call `CLI.run`. I've pushed a commit to change `CLI` to `DPPM::CLI`.

j8r:
Please `raise` instead of `abort`, the exception is then catched with a nice colored `ERR! this is my error` logging.

src/cli.cr
dscottboggs: I see. Ok.

src/cli.cr
```diff
@@ -2,7 +2,80 @@ require "dppm/cli"
 require "./dppm_rest_api"
 
 module DppmRestApi::CLI
-  def self.run_server(config, host, port, data_dir, **args)
+  extend self
+
+  # A helper method for the select_users method -- splits a comma-separated
+  # list of numbers in a string into an `Array` of `Int32` values.
+  private def split_groups_arg(argument text : String) : Array(Int32)
```
j8r: Why having space separated `argument text`? That's confusing. `numbers_list` is better.

src/cli.cr
dscottboggs: woaah. I did not know that.

src/cli.cr
```diff
@@ -2,7 +2,80 @@ require "dppm/cli"
...
+    text.split(',').map do |g|
```
j8r: That's a detail, this is creating an intermediate array.
Using the block variant of `split`, then concatenating with `numbers << g_int` will be more efficient.

src/cli.cr
dscottboggs: woaah. I did not know that.

> Please `raise` instead of `abort`, the exception is then catched with a nice colored `ERR! this is my error` logging.

Ah. :+1:

src/cli.cr
j8r: what is `g`? `raw_group_id` is a better name.

src/cli.cr
dscottboggs: agreed for sure

src/cli.cr
dscottboggs: Thanks!

src/config.cr
```diff
@@ -32,6 +32,25 @@ struct DppmRestApi::Config
...
+  def write_to(path : Path)
+    prefix = path.basename
+    suffix = path.extension
+    tmp_file = File.tempfile prefix, suffix, dir: path.dirname do |file|
+      self.to_json file
+    end
+    # to prevent data loss in case of loss of power during the file write.
+    File.rename tmp_file.path, path.to_s
+  ensure
+    if tf = tmp_file
```
j8r: Not needed, this type of conditions are mainly needed when we deal with method calls, not variables

dscottboggs: I'm sorry, which isn't needed? `if tf = tmp_file`?

j8r: The assignement isn't needed, sorry. Just `if tmp_file`.

dscottboggs: See, I thought that was the case, but someone in the chat said that there were circumstances where tmp_file could be non-nil in the conditional but then be nil when it gets to the call site, so you should just always assign to an intermediate variable. I just took their word for it -- no harm in doing it anyway, and I don't like `not_nil!` everywhere either.

Not that I'm saying I should keep it like this, just explaining why I did that -- it's just a habit I've developed. Now that you've brought it to my attention, though, I think `tmp_file.try &.delete` is better overall.

src/cli.cr
dscottboggs: This breaks things. See the difference between [8aac32c2 and c1c4e02](https://github.com/DFabric/dppm-rest-api/pull/8/commits/8aac32c2718b9ae0eb8c3f508386288774de3abd). I think that's an issue with DPPM's not being namespaced?

dscottboggs: nevermind it works after running shards update

dscottboggs: Actually it doesn't.

`action: :run_server` gives `undefined method 'run_server' for DPPM::CLI:Module`

`action: run_server` or `action: DppmRestApi::CLI.run_server` gives `undefined macro method 'Call#split'`

src/cli.cr
```diff
@@ -2,7 +2,247 @@ require "dppm/cli"
...
+                      API key that you already know, use the rekey command",
+            },
+          },
+          commands: {
+            edit: {
+              info:      "Edit one or more user's groups",
+              action:    "DppmRestApi::CLI.edit_users",
+              alias:     "e",
+              variables: {
+                new_name:   {info: "The name to change the selected user's name to"},
```
j8r: Can be split to a multiline, like others

src/cli.cr
```diff
@@ -2,7 +2,247 @@ require "dppm/cli"
...
+                },
+                key: {
+                  info: "the query key to match for",
+                },
+                add_glob: {
+                  info: "A glob to add to the allowed globs on this path",
+                },
+                remove_glob: {
+                  short: "rm",
+                  info:  "\
```
j8r: Why the backslash?

src/cli.cr
```diff
@@ -2,7 +2,247 @@ require "dppm/cli"
...
+                      A glob to remove from the allowed globs on this path",
+                },
+              },
+            },
+            delete: {
+              info:      "Delete the given group from the system",
+              alias:     "rm",
+              action:    "DppmRestApi::CLI.delete_group",
+              variables: {
+                group_id: {info: "The ID of the group to remove"},
```
j8r: Ditto, multiline

src/cli.cr
```diff
@@ -2,7 +2,247 @@ require "dppm/cli"
...
+        # find users by exact match
+        selected_users = selected_users.select do |user|
+          user.name == name
+        end
+      end
+    end
+    # Filter by group membership
+    if groups = match_groups
+      group_matches = Set(Set(Int32)).new
+      groups.split(':').each do |part|
```
j8r: You can use the block variant directly

j8r (commented):
Looks good overall, missing a few tests

dscottboggs:
> missing a few tests

Good point!

src/config.cr
```diff
@@ -1,20 +1,21 @@
...
 require "./ext/scrypt_password"
+require "json"
 
 struct DppmRestApi::Config
   include JSON::Serializable
   property groups : Array(Group)
   property users : Array(User)
 
-  def user(*, named username : String) : User?
+  def user_named(username : String) : User?
```
j8r:
```suggestion
  def user_from_name(username : String) : User?
```

src/config.cr
```diff
@@ -1,20 +1,21 @@
...
     users.find { |user| user.name == username }
   end
 
-  def user(*, authenticated_with api_key : String)
+  def user_authenticated_with(api_key : String)
```
j8r:
```suggestion
  def authenticated_user_from_api_key(api_key : String)
```

src/config.cr
```diff
@@ -1,20 +1,21 @@
...
     users.find { |user| user.api_key_hash == api_key }
   end
 
-  def group(*, named group_name : String) : Group?
+  def group_named(group_name : String) : Group?
```
j8r:
```suggestion
  def group_from_name(group_name : String) : Group?
```

src/config.cr
j8r: A bit long, if you have a better idea be free.

test.cr
```diff
@@ -0,0 +1,8 @@
+name = if true
+         ""
+       else
+         nil
+       end
+puts typeof(name) # => (String | Nil)
+name || abort "that's nil!"
+puts typeof(name) # => String
```
j8r: ?

dscottboggs: It was never here *waves hand*
I had copy-pasted that snippet into a file to run it, and it must've accidentally gotten committed when I was syncing my work from one computer to another

j8r: I do this sometimes too

spec/permissions.json
```diff
@@ -1,66 +1,80 @@
-
 {
   "groups": [
     {
       "name": "super user",
       "id": 0,
       "permissions": {
         "/**": {
-          "permissions": "All"
+          "permissions": 15
```
j8r: What are all this `15`? It was nice before.

src/config.cr
```diff
@@ -1,20 +1,21 @@
...
   property groups : Array(Group)
   property users : Array(User)
 
-  def user(*, named username : String) : User?
+  def user_from_name(username : String) : User?
     users.find { |user| user.name == username }
   end
 
-  def user(*, authenticated_with api_key : String)
+  def user_authenticated_by(api_key : String)
```
j8r: If we have a stack trace, with `user_authenticated_by` won't give us any clue what is it.
```suggestion
  def authenticated_user_from_api_key(api_key : String)
```
A bit long, but is it really a problem? Not sure we will use it that often.

spec/permissions.json
dscottboggs: It must be the default to serialize them as the numeric value, I can look into changing that.

src/config.cr
dscottboggs: It's not that it was a problem I just had done that before you suggested your version. I'll change it now so I don't forget.

dscottboggs: Actually, the *only* place where it even *would* be used... it [isn't](https://github.com/DFabric/dppm-rest-api/blob/b8b514f40038009c177c5dab8279bce21f41a339/src/middlewares/auth_middleware.cr#L34). I think we should just remove these methods.

dscottboggs: So I realize it's not working right now, but I wanted to get the idea fleshed out and see what you think. I think my Kemal Auth library can actually be reworked to be simpler, making use of modules as interfaces rather than needing a generic to be defined. I think I'm going to look into that before going on.

shard.yml
```diff
@@ -12,7 +12,8 @@ dependencies:
     github: kemalcr/spec-kemal
     branch: master
   scrypt:
-    github: ysbaddaden/scrypt-crystal
+    github: dscottboggs/scrypt-crystal
```
j8r, line 15: Have you done a PR for this?

spec/actions/app_spec.cr
```diff
@@ -28,37 +28,37 @@ module DppmRestApi::Actions::App
     end
     describe (fmt_route "/:app_name/service/boot") do
       pending "responds with 401 Forbidden" do
-        put (pp! fmt_route "/some_app/service/boot")
+        put (fmt_route "/some_app/service/boot")
```
j8r, line 31: Please, focus on the PR title to keep changes as minimal as possible - integrate the CLI.

run-tests.sh
```diff
@@ -0,0 +1,22 @@
...
+
+set -e
+
+build=no
+if [ "$1" = 'and' ] && [ "$2" = 'build' ]; then
+  shift; shift
+  build=yes
+fi
+
+set -e
```
j8r: Duplicated

run-tests.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+set -e
```
j8r: I also often add `-u` (`set -eu`) to ensure all variables are defined

run-tests.sh
```diff
@@ -0,0 +1,22 @@
...
+
+build=no
+if [ "$1" = 'and' ] && [ "$2" = 'build' ]; then
```
j8r: What is `and`?

spec/permissions.json
```diff
@@ -1,66 +1,80 @@
...
 {
   "groups": [
     {
       "name": "super user",
       "id": 0,
       "permissions": {
         "/**": {
-          "permissions": "All"
+          "permissions": "Create | Read | Update | Delete"
         }
```
j8r, line 14: Why this and not a list? This is a list of permissions.

spec/normal_user.api_key
```diff
@@ -0,0 +1 @@
+kP/U4BBQV0s2TcwN2MWwFrbwGDyMGkZEFt0era1yJc9eSuBhC+PiaDF1L+eylxsEoOzrnIsd4LgYOGQcIP6Y
```
j8r: Not sure it's needed to store this in a file.

run-tests.sh
dscottboggs: I meant to have `set -e` in one place, and `set -u` in another, to allow `$1` to be empty. Thanks for catching this

run-tests.sh
dscottboggs: `$  ./run-tests and build`

Just being clever. I'll cut it down.

spec/normal_user.api_key
dscottboggs: Should it just be generated on the fly as a constant then?

spec/permissions.json
dscottboggs: The `Access` type can be serialized as any of

- its numeric value,
- the above single-string form, or
- an array of string values.

The reserialization can be any of the above. There's no particular reason I picked the above form, it's just the first form that accepts multiples when you're reading it back. If you prefer the array form it can be that.

spec/normal_user.api_key
j8r: It will be better than reading to a file

spec/permissions.json
j8r: I think it's better be as close as standard json as possible, especially if  to be able to use external tools (`jq`...)

j8r: I think it's better to be as close as standard json as possible, especially if  to be able to use external tools (`jq`...)

j8r: I think it's better to be as close as standard json as possible, especially to be able to use external tools (`jq`...)

run-tests.sh
j8r: That can be an option, `--build`.
I guess this script will be used in the CI, too?

run-tests.sh
j8r: A trick is to use `${1-}` to bypass this.

run-tests.sh
dscottboggs: It doesn't need to be, it's just there for my own convenience.

run-tests.sh
dscottboggs: Cool

dscottboggs:
Can we merge this?

shard.yml
dscottboggs: https://github.com/ysbaddaden/scrypt-crystal/pull/6

j8r:
The CI will tell, it is red.

dscottboggs:
Yes, I need to merge the changes to `Scrypt::Password` into this and the JWT auth shard, then it should be ok. Is doing a separate branch/PR for that, then rebasing the right thing to do?

j8r:
Using your fork was fine. It's temporary and not perfect, but at least everyone can compile the project.

dscottboggs:
Ugh if I did that I'd have to wait for that to be merged and it won't even fix every instance. I'll just fix it on this branch and in the future I will create smaller-scale and more atomic PRs

dscottboggs:
Review is required for merging

run-tests.sh
```diff
@@ -0,0 +1,21 @@
...
+  printf "building...\r"
+  shards build > /dev/null
+  printf "building...done\n"
+}
+
+crystal tool format
+ameba
+KEMAL_ENV=test crystal spec $@
+[ $build = yes ] && do_build
+echo All OK
```
j8r: Missing new line

j8r (approved):

PR #10, opened 2019-05-26
Feature/package actions

dscottboggs:
Route definitions for pkg actions


src/utils.cr
```diff
@@ -20,6 +22,15 @@ macro fmt_route(route = "", namespace = false)
        .gsub(/::/, "/") }} {% if namespace %} + "/:namespace" {% end %} + ( {{route}} || "")
 end
 
+# Pull the context from the context's query parameter or use the default
+# provided by the CLI.
+#
+# This macro only works inside of a route block with the variable "context"
+# available
+macro prefix
+  context.params.query["prefix"]? || DPPM.default_prefix
```
j8r: A macro for this?!  They shouldn't be abused (slower compile time, harder to debug). Use a method taking a context as argument.

src/actions/pkg.cr
```diff
@@ -2,26 +2,114 @@ module DppmRestApi::Actions::Pkg
   extend self
   ALL_PKGS = ""
   ONE_PKG  = "/:id"
+
+  def output_config_for(package, to builder)
```
j8r: Do like in the stdlib https://crystal-lang.org/api/master/JSON/Serializable.html#to_json%28json%3AJSON%3A%3ABuilder%29-instance-method for the typing. (`json : JSON::Builder`)
I suggest a better self-explanatory method name, what does this method do?
I guess `build_package_vars`.

src/actions/pkg.cr
```diff
@@ -2,26 +2,114 @@ module DppmRestApi::Actions::Pkg
...
+    builder.field package.package do
+      builder.object do
+        package.pkg_file.config_vars.try &.each_key do |key|
+          builder.field key, package.get_config key
```
j8r, line 11: Use https://github.com/DFabric/dppm/blob/master/src/prefix/app.cr#L240 instead, else you'll miss library conf files like `/conf/php-fpm.ini`

j8r (changes requested):

src/utils.cr
j8r: It can also be moved out of utils, for example in `Actions` or in `DppmRestApi`

src/actions/pkg.cr
```diff
@@ -2,49 +2,172 @@ module DppmRestApi::Actions::Pkg
...
+        if key = ctx.params.query["get"]?
+          # A key was specified by the &get query parameter
+          begin
+            {
+              package_name => {
+                key => selected_pkg.get_config key,
+                # This method call  ^^ raises an untyped error if the key is not found
+              },
+              "errors" => [] of Nil,
+            }.to_json ctx.response
```
j8r: `JSON.build` can be used

src/actions/pkg.cr
```diff
@@ -2,49 +2,172 @@ module DppmRestApi::Actions::Pkg
...
+            ctx.response.flush
+            next ctx
+          rescue e : Exception
+            # catch the exception thown by the lack of a config value for the specified key
+            # otherwise, pass it on to the
+            msg = e.message || raise e
+            raise e unless msg.starts_with? "config key not found"
```
j8r: No need to do this. `get_config` has a block variant that yields when the key isn't found https://github.com/DFabric/dppm/blob/master/src/prefix/app.cr#L197
I'm creating a custom error anyway. Not sure what option is the best. On one hand using an exception brings a standard message, on the other hand the block one will be faster and more elegant.

src/utils.cr
```diff
@@ -1,8 +1,9 @@
-def deny_access!(to context)
-  context.response.status_code = 401
-  context.response.write "Forbidden.".to_slice
-  context.response.flush
-  context
+require "dppm/prefix"
+
+def deny_access!(to ctx)
```
j8r: Its place is in the `Actions` module.

src/actions/pkg.cr
dscottboggs: > what does this method do?

It *outputs* the *config for package* into *builder*. I don't know how it could be more clear than that. Although, I agree more type restrictions are better.

dscottboggs: I guess `build_config_response_for`? I suppose *output* isn't the right word.

src/actions/pkg.cr
dscottboggs: I don't understand, this is for the package. The application configurations are a different route. Or am I misunderstanding how the app/pkg structure works?

j8r: Yes sorry, I meant using `Pkg#each_config_key`, else dependencies configuration will be missed (php-fpm, in the future JVM).

dscottboggs:
Ok that's what I was thinking. Thanks.

src/actions/pkg.cr
j8r: You can also now catch `ConfigKeyError`

src/actions/pkg.cr
```diff
@@ -1,44 +1,146 @@
...
+          # No specific key was requested, respond with all of the config options
+          # for this package and all dependent packages
+          JSON.build context.response do |builder|
+            builder.object do
+              output_config_for selected_pkg, to: builder
+              if context.params.query["deps"]? && (deps = selected_pkg.pkg_file.deps)
+                builder.field "dependencies" do
+                  deps.each do |name, version|
+                    semver = SemanticVersion.parse version
+                    DPPM::Prefix.new(prefix).each_pkg do |pkg|
```
j8r: Why iterating over all packages?

j8r: "libraries" or "installed libraries" is a more correct word, because there are optional dependencies like `git` (for the filebrowser application).
You can use `Pkg#libs`.

src/actions/pkg.cr
```diff
@@ -1,44 +1,146 @@
...
+      builder.object do
+        package.pkg_file.config_vars.try &.each_key do |key|
+          builder.field key, package.get_config key
+        end
+      end
+    end
+  end
+
+  def find_package_by_name(name, pfx) : DPPM::Prefix::Pkg?
+    DPPM::Prefix.new(pfx).each_pkg { |pkg| return pkg if pkg.package == name }
```
j8r: Instantiate directly the package, then checking if it exists.
Remember that a package name is `mypackage_1.0.0`.
Else, if we want info for `mypackage`, that's the source one (`Src`)

src/actions/pkg.cr
j8r: Actually this method won't build a config response _for_ a package, but from, and for the client. I think better to drop this `_for` suffix.

dscottboggs: Ok.

src/ext/dppm_pkg.cr
```diff
@@ -0,0 +1,55 @@
...
+
+  def serialize_configuration(key : String, into builder : JSON::Builder)
+    errors = [] of String
+    builder.object do
+      builder.field package do
+        builder.object do
+          builder.field key do
+            builder.string value: get_config key
+          rescue e
+            raise e unless e.starts_with? "config key not found"
```
j8r: Please use `ConfigKeyError`

src/actions/pkg.cr
```diff
@@ -1,45 +1,137 @@
...
+        selected_pkg = DPPM::Prefix::Pkg.create get_prefix_or_default(from: context),
+          package_name,
+          version,
+          tag
+        raise NoSuchPackage.new context, package_name if selected_pkg.nil?
+        # Stop here ^^ unless the package whose name was the :id URL parameter
+        # was found and we can query it
+        if key = context.params.query["get"]?
+          data = selected_pkg.get_config? key
+          if data.nil?
```
j8r: Use `ConfigKeyError`.
A config key can be present, with a nil value.

src/actions/pkg.cr
```diff
@@ -1,45 +1,137 @@
...
-    if context.current_user? && Config.has_access? context, Access::Create
-      # TODO: build the package based on the submitted configuration
+    # Query information about a given package
+    relative_get "/:id/query" do |context|
+      if context.current_user? && Config.has_access? context, Access::Read
+        package_name = URI.unescape context.params.url["id"]
+        version = context.params.query["version"]?.try { |v| URI.unescape v }
+        tag = context.params.query["tag"]?.try { |t| URI.unescape t }
+        # Iterate over the packages to find the relevant one.
+        selected_pkg = DPPM::Prefix::Pkg.create get_prefix_or_default(from: context),
```
j8r: We don't want to create a new package, it already exists.
You can use `Prefix#new_pkg`.
I'll improve the documentation on this point.

src/actions/pkg.cr
```diff
@@ -1,45 +1,137 @@
...
+        end
+        next context
+      end
+      raise Unauthorized.new context
+    end
+    # Delete a given package
+    relative_delete "/:id/delete" do |context|
+      if context.current_user? && Config.has_access? context, Access::Delete
+        package_name = URI.unescape context.params.url["id"]
+        selected_pkg = DPPM::Prefix::Pkg.create get_prefix_or_default(from: context),
```
j8r: Same here, the package is supposed to already exist. Use `Prefix#new_pkg`.

src/actions/utils.cr
```diff
@@ -0,0 +1,55 @@
...
+      end
+    end
+    {% end %}
+
+    # Pull the context from the context's query parameter or use the default
+    # provided by the CLI.
+    #
+    # This macro only works inside of a route block with the variable "context"
+    # available
+    macro get_prefix_or_default(from context)
```
j8r: Use a `def`

j8r (changes requested):

src/actions/pkg.cr
j8r: Maybe I can even delete `Pkg.create`, I'm working on that.

src/actions/pkg.cr
dscottboggs: oof good catch, thank you

src/actions/pkg.cr
dscottboggs: Hm. Yeah, if the package is supposed to exist already, I don't think `new` or `create` are the best choices. Perhaps `load` or something like that?

src/actions/pkg.cr
```diff
@@ -1,147 +1,129 @@
...
-      if selected_pkg = find_package_by_name package_name, prefix
-        # The package whose name was the :id URL parameter was found and we can query it
+    # Query information about a given package
+    relative_get "/:id/query" do |context|
+      if context.current_user? && Actions.has_access? context, Access::Read
+        package_name = URI.unescape context.params.url["id"]
+        version = context.params.query["version"]?.try { |v| URI.unescape v }
+        # Iterate over the packages to find the relevant one.
+        selected_pkg = get_prefix_or_default(from: context).new_pkg package_name, version
+        raise NoSuchPackage.new context, package_name if selected_pkg.nil?
```
dscottboggs: Oh, `new_pkg` and `Pkg.create` don't return `Pkg?` so how will I know if the package hasn't been built?

src/actions/pkg.cr
j8r: The problem is the `create` method. We should only use `Prefix` to create new package objects, existing or not.
Then, we can use `#exists?` or `#exists!` to check/ensure its presence.

dscottboggs: Ok excellent.

src/actions/pkg.cr
j8r: `Pkg` is package object, a virtual package if you prefer.
I may create `VirtualPkg` and `Pkg`, not done yet. For now, you can use `#exists?` and `#exists!` (`program_data.cr`)

dscottboggs:
The problem is that [this line](https://github.com/DFabric/dppm-rest-api/blob/39cf0c8fc1aae8d247a32b2956276bf02914a561/src/actions/pkg.cr#L44) checks for `context@current_user` before calling `has_access?` -- which we don't have an easy way to set up. I'll have to change it so that the `context.current_user?` check is in the `access_filter` proc

(this line)
```crystal
      if context.current_user? && Actions.has_access? context, Access::Delete
```

j8r:
I've removed `Pkg.create` in the latest version of `DPPM` - much nicer now.

src/actions/utils.cr
```diff
@@ -0,0 +1,76 @@
...
+    # attempts to guard against that mistake by making `false`, `0`and `no`
+    # (case insensitive) falsey for our purposes. That is, this method returns
+    # `false` if the key is not present, or in the case of something like
+    # `?write=0`. In any other case it returns `true`.
+    #
+    # Anyone who is aware of this common mistake will of course instead use
+    # the proper boolean format, like `?value&other`
+    def parse_boolean_param(key : String, from context : HTTP::Server::Context) : Bool
+      if param = context.params.query[key]?
+        return false unless ["false", "no", "0"].includes? param.downcase
```
j8r:
```suggestion
        return false unless {"false", "no", "0"}.includes? param.downcase
```
Why we also support `yes` and `no`?
Can you provide some documentation for this like MDN?

src/actions/app.cr
```diff
@@ -1,185 +1,188 @@
...
+        next context
+      end
+      raise Unauthorized.new context
+    end
+    # start the service associated with the given application
+    relative_put "/:app_name/service/boot" do |context|
+      if context.current_user? && Actions.has_access? context, Access::Update
+        # TODO: boot the service
+        next context
+        puts "user found!"
```
j8r: Do not forget to remove.

spec/spec_helper.cr
```diff
@@ -25,13 +25,15 @@ macro assert_unauthorized(response)
...
-# after yielding to the block.
-def without_authentication!
-  DppmRestApi::Actions.access_filter = ->(_context, _permissions) { true }
-  yield
-  DppmRestApi::Actions.access_filter = DppmRestApi.access_filter
+module DppmRestApi
+  # Disable all authentication checks by making `DppmRestApi::Actions.has_access?`
+  # always return true, then set the authentication back to the default system
+  # after yielding to the block.
+  def self.without_authentication!
```
j8r: Not a good idea to monkey patch it: by actions reading the specs, we may think it's available in the library, but it is only for testing.

j8r: Not a good idea to monkey patch it: by reading the actions specs, we may think it's available in the library, but it is only for testing.

dscottboggs: The property is private. I had method next to this which called this method, so that it would be at the global namespace, but I thought you wouldn't like the extra abstraction. We could also make the property not protected.

src/actions/utils.cr
dscottboggs: I can remove "no", I agree that that's silly.

spec/spec_helper.cr
j8r: I deleted my comment by mistake because I thought I have submitted it twice :/ (I said it wasn't a good idea to monkey patch be cause the method is only used in specs).

I don't understand, have you an example to show why a normal top_level method can't be used?

src/errors/no_such_pkg.cr
```diff
@@ -0,0 +1,9 @@
+module DppmRestApi
+  class NoSuchPackage < NotFound
+    setter cause
+
+    def initialize(context, package_name : String)
+      super context, "no such message found called \"" + package_name + '"'
```
j8r:
```suggestion
      super context, "no such package found called \"" + package_name + '"'
```

src/actions.cr
```diff
@@ -1,38 +1,21 @@
...
-macro relative_{{method.id}}(route, &block)
-  {{method.id}} fmt_route(\{{route}}) do |\{{block.args.splat}}|
-    namespace = \{{block.args[-1]}}.params.query["namespace"]? || DPPM::Prefix.default_group
-    \{{block.body}}
-  end
-end
-{% end %}
-
 require "./access"
+require "./actions/utils"
```
j8r: It would be more logical to include `utils` where it is used, in each action file.

src/actions/pkg.cr
```diff
@@ -1,45 +1,129 @@
...
+    # a result on completion.
+    #
+    # This route takes the optional query parameters "version" and "tag".
+    relative_post "/:id/build" do |context|
+      if Actions.has_access? context, Access::Create
+        package_name = URI.unescape context.params.url["package"]
+        pkg = get_prefix_or_default(from: context).new_pkg package_name,
+          version: context.params.query["version"]?
+        pkg.build confirmation: false { }
+        pkg.exists!
```
j8r: I raises if the package exists, the method name is confusing. I should be `available!`
~Anyway, that's the library job to do this checks.~

spec/spec_helper.cr
dscottboggs: in the commit I just made -- `protected method 'access_filter=' called for DppmRestApi::Actions:Module`

src/actions/utils.cr
dscottboggs: Also I don't know where I got the idea that boolean parameters were standardized to be `?value=false` being truthy. That seems to just be wrong.

spec/spec_helper.cr
```diff
@@ -43,6 +41,12 @@ Fixtures.reset_config
 # Run the server
 DppmRestApi.run Socket::IPAddress::LOOPBACK, DPPM::Prefix.default_dppm_config.port, __DIR__
 # Set all configs back to the expected values, in case they changed
-Spec.before_each { Fixtures.reset_config }
+Spec.before_each do
+  Fixtures.reset_config
+  FileUtils.mkdir_p Fixtures::PREFIX_PATH
```
dscottboggs: Is there a way for DPPM to instantiate the directory structure that it needs? I'm getting a *no such file or directory* error when trying to clean packages on an empty prefix.

src/actions/pkg.cr
j8r: Currently, a package that already exists won't return an error, but it will for an application.

spec/spec_helper.cr
j8r: Yes, `Prefix#create`

j8r:
Do not forget to `shards update` for the get latest DPPM commit :wink:

src/actions.cr
dscottboggs: Utils are inherently tools that are there for all actions, so it makes more sense to me for it to be here, but if you feel strongly about it I can change it. It all goes in to the same tree anyway.

src/actions/pkg.cr
dscottboggs: ok. Is there any indication of failure for that method? Oh... I should be raising an `HTTPStatusError` from that block, shouldn't I?

j8r: What block? This method (now named `available!`) will raise an exception

spec/fixtures.cr
```diff
@@ -12,6 +12,8 @@ module Fixtures
     NORMAL_USER = Random::Secure.base64(TEST_KEY_SIZE)
   end
 
+  PREFIX_PATH = Path[__DIR__, "prefix"].to_s
```
j8r:
```suggestion
  PREFIX_PATH = File.tempname("_temp_dppm_prefix")
```

src/actions/utils.cr
```diff
@@ -0,0 +1,79 @@
...
+    # ...which is not valid JSON due to the unclosed object written at the beginning.
+    def build_json(response : IO)
+      JSON.build response do |json|
+        json.object do
+          json.field("data") { json.object { yield json } }
+        end
+      end
+    end
+
+    def ensure_dir_exists(dir : String)
```
j8r: This method shouldn't exist. All file_system operations should stay in the DPPM library.
Furthermore, the DPPM prefix should be already created when running the webserver, no need to check it anymore.

j8r (commented):
IMHO the prefix doesn't have to be configurable.
It would be safer, and also I don't see a use case since we have namespaces.

src/actions.cr
j8r: If that's for all actions, it make sense.
So, I don't think the utils name is appropriate, maybe `RouteHelper`

src/actions/pkg.cr
dscottboggs: I meant from `#build`, sorry

src/actions.cr
dscottboggs: Sounds good.

dscottboggs:
> IMHO the prefix doesn't have to be configurable.

That's a fair point. However, that's a very major change and would involve updates to the CLI and `Config` types, so I think it would be easier to make that a separate PR and continue with this as though that weren't going to happen.

src/actions/pkg.cr
j8r: Yes of course, there are lots of scenario for a build to fail.
The blocks is yield just before any change is done on the file-system. With the use of a boolean, we can choose to return a `400`, then after the block a `500`. What do you think?

dscottboggs: > The blocks is yield just before any change is done on the file-system

Please document these inconsistent uses of block yielding. I would say that even yielding to the block on an error is unidiomatic and should be documented.

> With the use of a boolean, we can choose to return a 400, then after the block a 500.

I'm sorry I don't understand what you mean by this

j8r: This is documented

dscottboggs: Oh, I see... :man_facepalming:

j8r: See https://github.com/DFabric/dppm/blob/master/src/prefix/pkg.cr#L109
> Build the package. Yields a block before writing on disk. When confirmation is set, the block must be true to continue.

The first stage is retrieving version, ensuring the packge exists etc.
If it fails, that's a `400 Bad Request`
Then, the block yields, and the second stage is executed, that performs changes on disk.
In any case of error, this is a `500 Internal Server Error`, because the build was supposed to work.

dscottboggs: I see. :+1:

src/actions/utils.cr
```diff
@@ -0,0 +1,79 @@
...
+    end
+    {% end %}
+
+    # Pull the context from the context's query parameter or use the default
+    # provided by the CLI.
+    #
+    # This macro only works inside of a route block with the variable "context"
+    # available
+    def get_prefix_or_default(from context)
+      DPPM::Prefix.new ensure_dir_exists context.params.query["prefix"]? || DPPM.default_prefix
```
j8r: You can use `Prefix#create`, that ensures the directories exist. `ensure_dir_exists` can then be removed.
As you said, this is temporary.

src/actions/pkg.cr
dscottboggs: Ok...maybe I don't see. So... the block is yielded to if the file already exists. I get that. Then, you simulate a build, but only [`if confirmation`](https://github.com/DFabric/dppm/blob/6016bf9da7688dd02ce03557371cbd4eaacccfd4/src/prefix/program_data.cr#L86) (which `confirmation` is false in this case). When `confirmation` is false, you still yield to the block, though, just without [logging a bunch of stuff](https://github.com/DFabric/dppm/blob/6016bf9da7688dd02ce03557371cbd4eaacccfd4/src/prefix/program_data.cr#L87-L103).

It looks like,  as a library user, I have to *count* the number of times the block is called, and determine which error to throw based on said count?

I realize it's getting quite late in your timezone so we can put this off until tomorrow if you prefer.

spec/spec_helper.cr
```diff
@@ -25,13 +25,21 @@ macro assert_unauthorized(response)
...
 end
 
-# Disable all authentication checks by making `DppmRestApi::Actions.has_access?`
-# always return true, then set the authentication back to the default system
-# after yielding to the block.
+module DppmRestApi
+  # Disable all authentication checks by making `DppmRestApi::Actions.has_access?`
+  # always return true, then set the authentication back to the default system
+  # after yielding to the block.
+  def self.without_authentication!
```
j8r: After thinking more, I think it's would be fine to monkey patch if we explicitly state this is used for specs:
```cr
def DppmRestApi.spec_without_authentication!(&block)
  Actions.access_filter = ->(_context : HTTP::Server::Context, _permissions : Access) { true }
  yield
  Actions.access_filter = ->access_filter(HTTP::Server::Context, Access)
end
```

spec/spec_helper.cr
```diff
@@ -25,13 +25,21 @@ macro assert_unauthorized(response)
...
+  end
+end
+
+@[AlwaysInline]
 def without_authentication!
-  DppmRestApi::Actions.access_filter = ->(_context, _permissions) { true }
-  yield
-  DppmRestApi::Actions.access_filter = DppmRestApi.access_filter
+  DppmRestApi.without_authentication! { yield }
 end
```
j8r:
```suggestion
DPPM::Log.output = DPPM::Log.error = File.open File::NULL, "w"
```
I've move `Log` to the `DPPM::Log` namespace in a new commit

spec/spec_helper.cr
dscottboggs: I don't like the phrasing of that. What if we create a `SpecHelpers` module (like we have with `RouteHelpers`), so that it's `SpecHelpers.without_authentication!`?

j8r: That's also good, as long as the name "spec" or "test" is inside to show that it is only used in spec tests.

src/actions/pkg.cr
j8r: The block is always yielded, and only once.
So we can have:
```cr
build_init = false
pkg.build confirmation: false { build_init = true }
```

j8r: We should also support simulate, later. The DPPM library isn't ready for that yet.

j8r:
I don't think making the prefix not configurable means to change evertyhing.
Just make sure `DppmRestApi.run` take also a prefix as argument, then create a protected `class_property`  in `Actions`.

I think it's ok to have the prefix configurable in the CLI, mainly because  the`dppm` command can be available globally on the system, but not when running the web rest api server.
By default, the dppm application will manage its prefix where it is installed – nice!

j8r:
I don't think making the prefix not configurable means to change everything.
Just make sure `DppmRestApi.run` take also a prefix as argument, then create a protected `class_property`  in `Actions`.

I think it's ok to have the prefix configurable in the CLI, mainly because  the`dppm` command can be available globally on the system, but not when running the web rest api server.
By default, the dppm application will manage its prefix where it is installed – nice!

src/dppm_rest_api.cr
```diff
@@ -28,12 +28,14 @@ module DppmRestApi
...
     port : Int32,
     data_dir : String,
+    prefix : String = DppmRestApi.default_prefix,
     access_filter : Proc(HTTP::Server::Context, Access, Bool) = ->access_filter(HTTP::Server::Context, Access)
   )
     ::File.open Path[data_dir, PERMISSIONS_FILE] do |data|
       @@permissions_config = Config.from_json data
     end
 
+    DppmRestApi.permissions_config.prefix = DPPM::Prefix.new(prefix).tap &.create
```
j8r:
```suggestion
    DppmRestApi.permissions_config.prefix = DPPM::Prefix.new prefix
```
Bad idea, I don't think we should create prefixes automagically.
The user have to explicitly do `dppm app install`. This can create a system service. This will be a way to start the API.
To check if the Prefix exists, use `Prefix#check`

src/config.cr
```diff
@@ -1,12 +1,25 @@
...
+  @_prefix : DPPM::Prefix? = nil
+
+  def prefix : DPPM::Prefix
+    @_prefix ||= DPPM::Prefix.new(@prefix).tap &.create
+  end
+
+  def prefix=(prefix : DPPM::Prefix)
+    @prefix = prefix.path.to_s
+    @_prefix = prefix
+  end
```
j8r: All of this isn't needed. Just `class_property : DPPM::Prefix { raise "no prefix set" }`.
The prefix don't have to be in the configuration, the sane default prefix managed should be where the `dppm` application is.

src/cli.cr
```diff
@@ -13,6 +13,10 @@ module DppmRestApi::CLI
           info:    "data directory",
           default: DppmRestApi::Actions::DEFAULT_DATA_DIR,
         },
+        prefix: {
+          info: "The base directory to install dppm packages into",
+          default: DPPM.default_prefix
+        }
```
j8r: No, it's already inherited.

j8r (dismissed):

src/cli.cr
dscottboggs: oh ok

src/dppm_rest_api.cr
```diff
@@ -11,7 +11,9 @@ module DppmRestApi
     raise "no permissions file is defined!"
   end
 
-  private def self.access_filter(context : HTTP::Server::Context, permission : Access) : Bool
+  class_property prefix : DPPM::Prefix { raise "no prefix set" }
```
j8r: That's a good idea to put it there :+1:

spec/permissions.json
```diff
@@ -0,0 +1,74 @@
...
+    {
+      "api_key_hash": "400$8$1e$10ba7fc8$0c8b2dabfd44ddb779d152d8cca0e4c7a695e85013a6a2f07c4c48b235986d85",
+      "group_ids": [
+        499,
+        1000
+      ],
+      "name": "Jim Oliver"
+    }
+  ]
+}
```
j8r: ?

scott-bossanova: wtf...?

dscottboggs: There is an exception being thrown when the tests are run, this file doesn't get cleaned up because of that exception

j8r: Why it is written there? It should be in the temp directory

dscottboggs: That's a very good question....

j8r: That's because of `Fixtures::PERMISSION_FILE`, it should use `File.tempname("_temp_dppm_permisions_file")`

j8r:
I suggest to move the `class_property prefix` to `Actions`. It's only used there AFAIK (the CLI will still have a configurable prefix).
The module will be more independent, too.

src/actions/pkg.cr
```diff
@@ -1,45 +1,142 @@
...
+    relative_post "/:id/build" do |context|
+      if Actions.has_access? context, Access::Create
+        package_name = URI.unescape context.params.url["id"]
+        pfx = DppmRestApi.prefix
+        pfx.update
+        pkg = pfx.new_pkg package_name, version: context.params.query["version"]?
+        init_done = false
+        begin
+          pkg.build confirmation: false { init_done = true }
+        rescue e
```
j8r: Nice! Little pedantic suggestion:
```cr
pkg.build confirmation: false do
  init_done = true
rescue e
...
end
```

src/actions/pkg.cr
dscottboggs: This feature doesn't seem to work according to my test? Please see [the latest CI results](https://cloud.drone.io/DFabric/dppm-rest-api/214/1/5) when you get a chance.

src/actions/pkg.cr
```diff
@@ -1,45 +1,142 @@
...
-    if context.current_user? && Actions.has_access? context, Access::Read
-      # TODO: Query information about the given package
-      next context
+
+    # List built packages
+    relative_get nil do |context|
+      if Actions.has_access? context, Access::Read
+        pkgs = [] of String
+        begin
+          DppmRestApi.prefix.each_pkg { |pkg| pkgs << pkg.package }
```
j8r: It shouls be a list of hashes, with at least the package name and the version. I think it's a sane default.
In the future (or now if you prefer), this elements will be configurable through parameters.
An user may wants also descriptions, documentation link, license etc.

src/actions/pkg.cr
```diff
@@ -1,45 +1,142 @@
...
+        rescue e
+          raise InternalServerError.new context,
+            message: "#each_pkg raised an exception with the message '#{e.message}'.",
```
j8r: This is an internal implementation detail, perhaps something more generic like "Package listing failed".
No need to add the exception message inside it, it will be shown already in the error trace.

src/actions/route_helpers.cr
```diff
@@ -0,0 +1,61 @@
...
+      end
+    end
+    {% end %}
+
+    # This method parses a boolean value from a query parameter. It returns
+    # true when the query parameter contains any string *except* for `"false"`
+    # or `"0"`.
+    def parse_boolean_param(key : String, from context : HTTP::Server::Context) : Bool
+      if param = context.params.query[key]?
+        return true unless {"false", "0"}.includes? param.downcase
```
j8r: Any reason I miss to support also `0`, in addition to `false`?

j8r (commented):
Good overall! Some notes on advanced features.

src/actions/route_helpers.cr
dscottboggs: Idk it just seems common to assume 1 and 0 will work for boolean operations.

j8r: Not really, we use `true` and `false` everywhere else (json, con, CLI, Crystal).
It seems also more common for other APIs to pass `true` and `false` (which are strings)

src/actions/pkg.cr
dscottboggs: This still gives an *Internal Server Error* instead of *Bad Request* when given a package name that does not exist. In fact, I don't see any scenario in which an exception will be raised before [the block is yielded](https://github.com/DFabric/dppm/blob/fa4faae67e475f87c820020854ce71c6309f6074/src/prefix/pkg.cr#L143-L150).

spec/permissions.json
dscottboggs: No, because the filename "permissions.json" is hardcoded, the server CLI only gets the directory that it's supposed to be stored in. In the case of the specs, that's `__DIR__`, which resolves to `$PROJECT_DIR/spec`.

src/actions/pkg.cr
dscottboggs: > this elements will be configurable through parameters. An user may wants also descriptions, documentation link, license etc.

A user will not be using the API directly. If the client wants to query that information about a package, they could use the *query* route.

src/actions/pkg.cr
j8r: Good point, sorry

spec/permissions.json
j8r: Not a problem, all the `data_dir: __DIR__` in `cli_spec` can be changed to the `Fixtures` path.

src/actions/pkg.cr
j8r: Why? This means when the Web client will list all applications, it will also have to do multiple requests?
We will have duplicate packages with this current implementation when we have two versions of the same package.

dscottboggs: Fair enough.

spec/permissions.json
dscottboggs: So... perhaps, instead of `__DIR__` and `Fixtures::PERMISSION_FILE`, we could have `Fixtures::DIR`? Furthermore, should this directory be within the project directory or within the system's tempdir?

src/actions/pkg.cr
```diff
@@ -1,45 +1,155 @@
...
+      def initialize(@version); end
+    end
+
+    # List built packages
+    relative_get nil do |context|
+      if Actions.has_access? context, Access::Read
+        pkgs = {} of String => ListResponse
+        DppmRestApi.prefix.each_pkg do |package|
+          if (found = pkgs[package.package]?) && found.version > package.semantic_version
+            # newer version of package already exists
```
j8r: Why does it matter? We want to list **all** packages.
I think the best would be an array of hash of string => string.

dscottboggs: OK so the structure of the response is like so

```json
{ "data": {
  "dppm": {
    "version": "2019.4.26"
  }
} }
```

This leaves room for expanding the API to have further information such as 

```json
{ "data": {
  "dppm": {
    "version": "2019.4.26",
    "license": "ISC",
    "description": "Dedicated Platform Package Manager - an easy way to install and manage server applications."
  }
} }
```
I thought you wanted me to remove duplicates in this method but now that I think about it that doesn't really make sense. I'll get rid of that part.

spec/actions/pkg_spec.cr
```diff
@@ -165,10 +165,10 @@ module DppmRestApi::Actions::Pkg
         BuildResponse.from_json(response.body).should_be_successful_for "dppm"
       end
     end
-    it "responds with Internal Server Error when given an invalid package name" do
+    it "responds with Bad Request when given an invalid package name" do
```
dscottboggs, line 167: This spec is failing due to an issue in the library I think.

src/actions/pkg.cr
j8r: What duplicates? With the examples you posted, it can't have several `dppm` packages, which can happen when upgrading for example.

j8r: I propose:
```json
"data":[
{
  "package": "dppm",
  "version": "1.2.3"
},
{
  "package": "dppm",
  "version": "1.2.7"
}
]
```
Of course it can also include description, license etc. as you said, depending of parameters.
It would be more or less a list of what's in `PkgFile`, plus the version.

src/actions/pkg.cr
```diff
@@ -1,45 +1,152 @@
...
-  end
-  # Delete a given package
-  relative_delete "/:id/delete" do |context|
-    if context.current_user? && Actions.has_access? context, Access::Delete
-      # TODO: Query information about the given package
-      next context
+
+    struct ListResponse
+      include JSON::Serializable
+      property version : SemanticVersion
```
j8r: All code relative to `SemanticVersion` isn't needed.
Also, what the `ListResponse` abstraction bring?
We will need similiar objects too for `app` and `src`.
The difference between the three will be:
- src: nothing, what's in pkgfile.
- pkg: add version
- app: add name (and perhaps version)

src/ext/semantic_version.cr
```diff
@@ -0,0 +1,27 @@
+require "semantic_version"
```
j8r: This file isn't needed.

src/dppm_rest_api.cr
j8r: I think it will be better in `Actions.prefix`, that were the prefix is used.

src/actions/pkg.cr
dscottboggs: It's not an abstraction, it's a type. We could use a NamedTuple if you're against having a defined struct...tbh I'm kind of confused as to what the issue is. The app and src versions can be `Actions::App::ListResponse` and `Actions::Src::ListResponse`.

j8r: Does JSON.build directly wouldn't be better? In the sense of being flexible for what fields we return to the client, like description, info, etc.
This will avoid having 3 types, and the logic can be mostly shared.

dscottboggs: But why? What makes JSON.build better? It would be more code, and more prone to mistakes.

j8r: How do you do to build only fields requested?

dscottboggs: Oh. That's a good point. If you want to do that, then yes, a `build` implementation would be more appropriate. If we are allowing bulk querying of config values, then I propose renaming this method to `/dppmrestapi/pkg/query`, and offer optional pagination on the response as a long-term goal once the project is more developed.

j8r: ... that's a good idea, I may implement this in the CLI at first.
Ok, we could let this as is and add a `TODO` comment.

spec/actions/pkg_spec.cr
j8r: The dppm library? I don't think so, it doesn't know the concept of API routes. I figured out why.

dscottboggs: The dppm library doesn't need to know API routes to raise an exception upon receiving an invalid package name. Which is what this spec is testing. Sorry, I was trying to point out [the body](https://github.com/DFabric/dppm-rest-api/blob/cfe755532ecbfb529208268df793ccd2b7482ef4/spec/actions/pkg_spec.cr#L170-L172), not which HTTP status was raised..

dscottboggs:
```crystal
post fmt_route "/$(echo all your base are belong to us!)/build"
# (which does)
pkg = pfx.new_pkg "$(echo all your base are belong to us!)", version: nil
# this fails both because there's no underscore in the package name
# with no version specified, and because the package name has invalid
# characters
init_done = false
pkg.build confirmation: false do
  # this block should never get yielded to, due to input validations raising
  # an exception before the block is called
  init_done = true
rescue e
  # which means this exception is raised.
  raise InternalServerError.new context, cause: e if init_done
  raise BadRequest.new context, cause: e
end
# making this line fail
response.status_code.should eq HTTP::Status::BAD_REQUEST.value
```

src/actions/pkg.cr
```diff
@@ -1,45 +1,152 @@
...
+        package_name = URI.unescape context.params.url["id"]
+        pfx = DppmRestApi.prefix
+        pfx.update
+        pkg = pfx.new_pkg package_name, version: context.params.query["version"]?
+        init_done = false
+        pkg.build confirmation: false do
+          init_done = true
+        rescue e
+          raise InternalServerError.new context, cause: e if init_done
+          raise BadRequest.new context, cause: e
```
j8r: This should be:
```cr
begin
  pkg = pfx.new_pkg package_name, version: context.params.query["version"]?
  pkg.build confirmation: false { init_done = true }
rescue ex
  raise InternalServerError.new context, cause: ex if init_done
  raise BadRequest.new context, cause: ex
end
```

spec/actions/pkg_spec.cr
j8r: This is due to some methods not being catched in `Actions::Pkg` (see my code suggestion).
A package can be built internally from a source (`Src` object) in order to take the latest version. If it doesn't exist, an exception is raised.
Setting the package name + version doesn't require to read a source package.

spec/spec_helper.cr
```diff
@@ -25,21 +25,33 @@ macro assert_unauthorized(response)
...
+    Actions.access_filter = ->(_context : HTTP::Server::Context, _permissions : DppmRestApi::Access) { true }
+    yield
+  ensure
+    Actions.access_filter = ->DppmRestApi.access_filter(HTTP::Server::Context, Access)
+  end
 end
 
+DPPM::Log.output = DPPM::Log.error = File.open File::NULL, "w"
 Kemal.config.env = "test"
+FileUtils.mkdir_p Fixtures::DIR.to_s
```
j8r, line 39:
```suggestion
FileUtils.mkdir_p Fixtures::DIR.to_s
at_exit { FileUtils.rm_rf Fixtures::DIR.to_s }
```
Add also `require "file_utils"`

spec/spec_helper.cr
```diff
@@ -25,21 +25,33 @@ macro assert_unauthorized(response)
...
 # Run the server
-DppmRestApi.run Socket::IPAddress::LOOPBACK, DPPM::Prefix.default_dppm_config.port, __DIR__
+DppmRestApi.run Socket::IPAddress::LOOPBACK,
+  DPPM::Prefix.default_dppm_config.port,
+  Fixtures::DIR,
+  DPPM::Prefix.new(Fixtures::PREFIX_PATH.to_s).tap &.create
 # Set all configs back to the expected values, in case they changed
-Spec.before_each { Fixtures.reset_config }
+Spec.before_each do
+  FileUtils.mkdir_p Fixtures::DIR.to_s
```
j8r: Already present
```suggestion

```

j8r:
Running spec is slow because each operation involve downloading `packages-source` archive and the `dppm` binary,  again and again.
I propose mocking instead.
Look at `./lib/dppm/spec/samples/testapp`
Edit: I can fix this in another PR, I guess.

j8r:
I have taken the initiative to push some commits. Be free to put comments on points you disagree.

src/actions/route_helpers.cr
j8r: The `param.downcase` means  that `FALSE` is also supported?!

j8r: I propose to only support `true` and `false`, else raise.

spec/actions/pkg_spec.cr
```diff
@@ -53,8 +53,7 @@ module DppmRestApi::Actions::Pkg
   end
 
   def build_test_package
-    DppmRestApi.prefix.update
-    pkg = DppmRestApi.prefix.new_pkg "dppm"
+    pkg = DppmRestApi.prefix.new_pkg "testapp"
```
dscottboggs: Why does this work?

spec/spec_helper.cr
```diff
@@ -49,6 +50,7 @@ Spec.before_each do
   Fixtures.reset_config
   FileUtils.mkdir_p Fixtures::PREFIX_PATH.to_s
   DppmRestApi.prefix.create
+  FileUtils.cp_r "./lib/dppm/spec/samples", DppmRestApi.prefix.src.to_s
```
dscottboggs: Oh, I see!! :100:

dscottboggs:
> I have taken the initiative to push some commits. Be free to put comments on points you disagree.

I was working on this process locally and this seems to fix all the issues I had and more. Thanks for your help! I'll get to some other tasks now.

src/actions/route_helpers.cr
```diff
@@ -15,6 +15,11 @@ module DppmRestApi::Actions
       {{method.id}} fmt_route(\{{route}}) do |\{{block.args.splat}}|
         namespace = \{{block.args[-1]}}.params.query["namespace"]? || DPPM::Prefix.default_group
         \{{block.body}}
+      rescue e
+        # Catch non-HTTPStatusError exceptions, and throw an InternalServerError
+        # with the original error as the cause.
+        raise e if e.is_a? HTTPStatusError
+        raise InternalServerError.new \{{block.args[0]}}, cause: e
```
j8r: I think it would be clearer to do something like:
```cr
...
rescue ex : Actions
  raise ex
rescue ex
  InternalServerError
end
```
This means moving exception dedicated to `Actions`, in the `Actions` module.

dscottboggs: Are you suggesting that [this](https://github.com/DFabric/dppm-rest-api/blob/cfe755532ecbfb529208268df793ccd2b7482ef4/src/errors/http_status_errors.cr#L5) be renamed to `Actions` or `ActionException` or something? Or perhaps an alias like `alias Actions = HTTPStatusError | SomethingElse`?

src/actions/route_helpers.cr
dscottboggs: I hadn't thought of raising `BadRequest`. Counter-proposal -- only support the existence or lack thereof and raise if there is *any* text given in a boolean parameter. `?libraries&something` would be the only option, anything else ( `?libaries=yes&something=false`) would raise `BadRequest`.

src/actions/pkg.cr
dscottboggs: What config values were you thinking of allowing in the parameters for this? I didn't see anything relevant in `Pkg` or `Base`, unless you were planning to allow querying arbitrary config values in this route?

j8r: Mainly what's in `PkgFille`, plus version for package and name for app.
This will help when we will build the Web client, listing all apps will require only one request.

src/actions/route_helpers.cr
j8r: I submitted a commit to show what I was thinking: having `Actions::Exception`.

src/actions/route_helpers.cr
j8r: Wait, why having this possibility?
I think I will rename `query` to `info`, because that's the info present in `pkg.con` file.
Querying the configuration is another endpoint, `/config`, and abstract the origin of the configuration.
The endpoint could be `/pkginfo`

src/actions/route_helpers.cr
```diff
@@ -1,68 +1,67 @@
...
-        raise InternalServerError.new \{{block.args[0]}}, cause: ex
-      end
+  # Transform a relative route to an absolute one from the `ROOT_PATH`.
+  def fmt_route(relative_route : String? = nil, namespace : Bool = false)
+    ROOT_PATH + (namespace ? "/:namespace" : "") + relative_route.to_s
+  end
+
+  {% for method in %w(get post put delete ws options) %}
+  # Perform a relative {{method.id}} from the `ROOT_PATH`.
+  def relative_{{method.id}}(route : String?, &block : HTTP::Server::Context ->)
```
dscottboggs, line 15: Why??

j8r: Because better to avoid unneeded macros. A classic method do the job.

dscottboggs: > better to avoid unneeded macros

What makes you say that? I don't see anything wrong with making a liberal use of macros. That said, after giving it some thought, in this case it does neaten things up a bit.

src/actions/route_helpers.cr
dscottboggs: Is this comment about a different section of code? I'm confused

src/actions/route_helpers.cr
j8r: I disagree, macros have strong downsides. That why they must not be abused, and **only** used when strongly needed.
- harder to debug, specially compile time errors
- slower compile times
- lazy typing, no type for arguments, no signature
- unwanted side effects (modifying variables in the current scope)

src/actions/pkg.cr
```diff
@@ -1,44 +1,150 @@
...
+    if Actions.has_access? context, Access::Read
+      JSON.build context.response do |json|
+        json.object do
+          json.field "data" do
+            json.array do
+              Actions.prefix.each_pkg do |package|
+                json.object do
+                  json.field "package", package.package
+                  json.field "version", package.version
+                  package.each_config_key do |key|
```
j8r: I don't really see the point for packages. However for applications it will be useful.

src/actions/pkg.cr
```diff
@@ -1,44 +1,150 @@
...
+                    if parse_boolean_param key, from: context
+                      json.field key do
+                        package.get_config(key) { nil }.to_json json
```
j8r: Use `get_config` instead, which will raise.
Note there is also `get_config?`

src/actions/pkg.cr
dscottboggs: so leave this off, and we're ready to merge?

src/actions/route_helpers.cr
```diff
@@ -0,0 +1,70 @@
...
+
+  # This method parses a boolean value from a query parameter. It returns
+  # true when the query parameter is specified, but empty; false when the
+  # parameter is not specified (null); and otherwise raises a BadRequest
+  # exception.
+  def parse_boolean_param(key : String, from context : HTTP::Server::Context) : Bool
+    context.params.query[key]?.try do |value|
+      raise BooleanParamHasValue.new context, key, value unless value == ""
+      return true
+    end
```
j8r, line 38: I propose something a bit more idiomatic:
```cr
if value = context.params.query[key]?
  return true if key.empty?
  raise BooleanParamHasValue.new context, key, value
end
false
```

src/actions/pkg.cr
dscottboggs: > Use get_config instead, which will raise.

Note that the current implementation of `build_json` requires that [no exceptions be thrown](https://github.com/DFabric/dppm-rest-api/blob/adbae12b5e71c925ff0a1404631902849a4c3385/src/actions/route_helpers.cr#L42-L62) while the response is being built. Until now, the only resolution to this I could think of was building to an intermediary IO and copying to `context.response` after it's built. That would be a big performance hit in a crucial spot. However, talking another look at my implementation I was thinking that there is one alternative, although it is less than ideal....suggestions are welcome, but I'd recommend we discuss this in a separate PR, so that I can also begin work on other routes while we work out the details.

```crystal
def build_json(response : IO)
  error : Exception? = nil
  JSON.build response do |json|
    json.object do
      json.field "data" do
        json.object do
          yield json
        rescue e
          error = e
        end
      end
      if error
        json.field "errors" do
          ErrorResponse.new(error).errors.to_json json
        end
      end
    end
  end
end
```

I realize this would require specifying a partial succesful response, but there's no way to rewind an `HTTP::Server::Response` because it's *streamed* out to the client on the fly.

dscottboggs: For the package routes, I (think) I successfully avoided cases which raise, so that's an option as well, but it's...fragile, and I don't like it, especially not long-term

src/actions/pkg.cr
```diff
@@ -1,44 +1,143 @@
...
+      package_name = URI.unescape context.params.url["id"]
+      version = context.params.query["version"]?.try { |v| URI.unescape v }
+      # Iterate over the packages to find the relevant one.
+      selected_pkg = Actions.prefix.new_pkg package_name, version
+      raise NoSuchPackage.new context, package_name unless selected_pkg.exists?
+      # Stop here ^^ unless the package whose name was the :id URL parameter
+      # was found and we can query it
+      if key = context.params.query["get"]?
+        data = begin
+          selected_pkg.get_config key
```
j8r:
```suggestion
         config_key = selected_pkg.get_config key
```
Then remove `data = begin`

src/actions/pkg.cr
j8r: Yes, we will have to figure out how to do it. We could have an abstraction for this and use it anywhere we need.
For a reason or another, building the JSON may result of an error.

PR #11, opened 2019-06-02
Bug/Authentication group not found

dscottboggs:
This fixes a bug where a user would be a member would not be recognized as authenticated in the following case:

1. The user is a member of a group for which the glob pattern matches the current request path, for which they **are not** allowed access, and
2. The user is also a member of a group for which the glob pattern matches the current request path, for which they **are** allowed access

This is the issue that was preventing most of the route specs from working, however, there is still an unrelated issue that is still preventing some of them from working


scripts/setup-for-manual-testing.sh
```diff
@@ -0,0 +1,43 @@
...
+dppm server group add "name=${DPPM_USER}s group" id=$GROUP_ID
+dppm server add_user "name=$DPPM_USER" groups=$GROUP_ID
+
+if [ $ADMIN = yes ]; then
+  make_admin
+fi
+
+./bin/dppm server run
+
+rm -r data
```
j8r: what is this file? It can rather be done as an integration spec

j8r (dismissed):

scripts/setup-for-manual-testing.sh
dscottboggs: It's a script for quickly setting up a version of the server for manual testing. I was doing some experimenting with curl and jq to try to figure out this and another issue.

j8r: Well, why not. Please make it POSIX, and we are good.

scripts/setup-for-manual-testing.sh
```diff
@@ -0,0 +1,43 @@
...
```
j8r: miss new line

dscottboggs: :roll_eyes: this is what I get for hopping editors

j8r (approved):

PR #13, opened 2019-06-10
refactor error handling -- option 1 of 2

dscottboggs:
I have been thinking about how to handle this refactor for the past few days, and have come up with two options that I like. I created a separate PR for each idea and I would like your opinion on which is better. The other is #14.

The key difference is that this version adds a list of errors directly to the context, while the other choice is more generic.


j8r:
Can you create an issue to track this both PR? With the problem explained (why 2 options, pros/cons, why not using standard exception etc.)

I don't quite like either solution, which both result of monkey patching the stdlib `HTTP::Server` with new methods, which is IMO avoidable.
Have you considered a custom error, or if stack unwinding worry you, a library like https://github.com/j8r/error.cr?

PR #12, opened 2019-06-02
bug/route matching

dscottboggs:
There is an issue with some of the routes being recognized and responding. I'm not sure if this is an issue with our API structure or with Kemal.

PR #16, opened 2019-06-10
Development fixes

j8r:
Various fixes related to development, to avoid having the CLI and macro debug printed.
We may have a debug flag, if it's useful to you.

Two problems remains:
- running `crystal spec` hangs indefinitely. Ideally, it should work out of the box.
- a hash in `spec/permissions.json` changes, this is a bit annoying. It may be linked to the upper command, so adding to `.gitignore` isn't a proper solution.

### DIFF
--- a/src/utils.cr
+++ b/src/utils.cr
@@ -18,7 +18,6 @@ macro fmt_route(route = "", namespace = false)
        .downcase
        .gsub(/^DppmRestApi::Actions::/, "")
        .gsub(/::/, "/") }} {% if namespace %} + "/:namespace" {% end %} + ( {{route}} || "")
-  {% debug %}
 end
 
 # the "last" or -1st block argument is selected because context is always the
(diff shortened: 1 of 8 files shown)


src/utils.cr
```diff
@@ -29,7 +28,6 @@ macro relative_{{method.id}}(route, &block)
     namespace = \{{block.args[-1]}}.params.query["namespace"]? || DPPM::Prefix.default_group
     \{{block.body}}
   end
-  \{% debug %}
   puts "built route at {{method.upcase.id}}:\t#{fmt_route(\{{route}})}"
```
dscottboggs: this line can go too -- they were there to track down #12, and I just hadn't gotten to cleaning them up yet. Thanks for taking care of that for me

dscottboggs (changes requested):

dscottboggs (approved):

j8r:
Thanks @dscottboggs (and perfect merge) :)

PR #18, opened 2019-06-11
Simplify specs running

### LINKED ISSUE #17: Running specs out of the box
- running `crystal spec` hangs indefinitely. Ideally, it should work out of the box.
- a hash in `spec/permissions.json` changes, this is a bit annoying. It may be linked to the upper command, so adding to `.gitignore` isn't a proper solution.

j8r:
Fixes #17  (but not the `permissions.json` file changed). I think `permissions.json` should be created programmatically, and put in a temporary location, which will be deleted.
I haven't used `Spec.before_each`, because it creates a new API server for each test, which increase time to run the suite - do we really want this?


spec/actions/app_spec.cr
```diff
@@ -1,3 +1,4 @@
+require "../../src/actions"
```
dscottboggs, line 1: Why add a require here, instead of somewhere within *spec_helper*?

spec/spec_helper.cr
```diff
@@ -1,20 +1,12 @@
...
-  Path[__DIR__, "permissions.json"]
-end
-
-def new_test_context(verb = "GET", path = "/api/test")
-  backing_io = IO::Memory.new
-  request = HTTP::Request.new verb, path
-  response = HTTP::Server::Response.new backing_io
-  {backing_io, HTTP::Server::Context.new(request, response)}
-end
+Kemal.config.env = "test"
```
dscottboggs, line 8: This works?? Awesome! :tada:

dscottboggs (dismissed):
Very nice, I'm really excited about being able to avoid prefixing my `crystal spec` commands with `KEMAL_ENV=test`. I have an idea about permissions.json -- ECR. I'll try to experiment with it tonight.

scripts/run-tests.sh
```diff
@@ -1,21 +0,0 @@
-#!/bin/sh
```
dscottboggs, line 1: why have you removed this?

dscottboggs:
> I have an idea about permissions.json -- ECR

I tried this -- not only (in my initial tests) did it not work, but required the use of a string literal instead of a constant (presumably a bug). Perhaps not an ideal solution, but it could work with some tweaking... will have to try again tomorrow.

spec/actions/app_spec.cr
j8r: This isn't the purpose of `spec_helper`.
We require only what we need, depending on what part we test.
Therefore, we can affirm that only this module of our app is tested, which is self-contained.

This also speed up running individual files. That's how it's done in the Crystal repo too.

scripts/run-tests.sh
j8r: Not needed, this can be replaced by a one liner (as specified in the README.md): `crystal tool format && bin/ameba && crystal spec`, then `shards build` can also be appended.

j8r:
ECR/templating doens't seem to be a good idea, either.
All should be possible with only `DppmRestApi::Config`, like for a library. That's what's this module designed to do in the first place.

j8r:
ECR/templating doens't seem to be a good idea, either.
All should be possible with only `DppmRestApi::Config`, like for a library. That's what's this module designed for in the first place.

dscottboggs (approved):
Please review my proposed solution to this issue, and modify as you see fit.

j8r:
What does fixtures mean? Is it really needed to monkey patch `DppmRestApi`? Other than that, I love the idea to create the config file programatically.

I think your commit can be another PR.
Let's validate this one without your work, merge it when you are OK, then review your work to improve the `permissions.json` case.

dscottboggs:
> What does fixtures mean?

I suppose it's more common in UI testing -- it's just what I'd seen it called before, if you have a different name for it that is more appropriate, that's fine. See https://stackoverflow.com/a/14684400 and https://en.wikipedia.org/wiki/Test_fixture#Software

> Is it really needed to monkey patch DppmRestApi?

Certainly not. I don't see it as monkey-patching, the whole project is the DppmRestApi module. Anything within the project is a part of the module. There's certainly no reason though that the Fixtures module can't be at the top-level.

> I think your commit can be another PR.

Sure, I'll rework it when I get the chance.

PR #19, opened 2019-06-16
add fixtures file in place of permissions.json

dscottboggs:


spec/spec_helper.cr
```diff
@@ -15,12 +12,26 @@ def new_test_context(verb = "GET", path = "/api/test")
...
-# need to update the configuration to match the key that was just generated.
-usr = DppmRestApi.permissions_config.users.find { |user| user.name == "Jim Oliver" }.not_nil!
-DppmRestApi.permissions_config.users.delete usr
-usr.api_key_hash = Scrypt::Password.create NORMAL_USER_API_KEY
-DppmRestApi.permissions_config.users << usr
-DppmRestApi.permissions_config.write_to PERMISSION_FILE
+# the location
+macro permissions_file!
+  Path[__DIR__, "permissions.json"]
+end
```
j8r: The previous `PERMISSION_FILE` constant was better, a macro isn't needed here.

spec/fixtures.cr
```diff
@@ -0,0 +1,63 @@
...
+          groups:       [0],
+          name:         "Administrator",
+        },
+        {
+          api_key_hash: Scrypt::Password.create(TEST_USER_RAW_API_KEYS[:normal_user]),
+          groups:       [499, 1000],
+          name:         "Jim Oliver",
+        },
+      ],
+    }
```
j8r: Better, but I think we can further improve this: How this named tuple was created? This is too magic. I suppose the `DppmRestApi` was used at first, all we should need is only `DppmRestApi::Config`.
We may also have a problems if the JSON shema change.

j8r (changes requested):

spec/fixtures.cr
```diff
@@ -0,0 +1,66 @@
+require "../src/config"
+require "../src/ext/scrypt_password"
+
+module DppmRestApi
+  struct Config
+    def initialize(@groups, @users)
```
j8r: Why? If this initializer is needed, other may need it too – so it can be added to `/src/config.cr`.

spec/fixtures.cr
```diff
@@ -0,0 +1,66 @@
+require "../src/config"
+require "../src/ext/scrypt_password"
```
j8r: Not needed

spec/spec_helper.cr
dscottboggs: Assigning this to a constant broke things elsewhere due to `Path.[]` being a `Call` as opposed to `StringInterpolation` in macro-land. However, that issue seems to have cleared up.

spec/spec_helper.cr
```diff
@@ -15,12 +12,19 @@ def new_test_context(verb = "GET", path = "/api/test")
...
-usr.api_key_hash = Scrypt::Password.create NORMAL_USER_API_KEY
-DppmRestApi.permissions_config.users << usr
-DppmRestApi.permissions_config.write_to PERMISSION_FILE
+# the location
+PERMISSION_FILE = Path[__DIR__, "permissions.json"]
+
+Spec.before_each do
+  DppmRestApi.permissions_config = Fixtures.permissions_config
+  DppmRestApi.permissions_config.write_to PERMISSION_FILE
+  DppmRestApi.run Socket::IPAddress::LOOPBACK, DPPM::Prefix.default_dppm_config.port, __DIR__
```
j8r, line 25: You think it's better to create a new server for each spec? This will slow it down.
Reseting the permission file may be enough, with eventually reloading the server.

j8r (approved):
Nice! One more question.

j8r:
For some reason this branch can't be built in the CI – it was stuck with a High GC allocation count warning.

spec/spec_helper.cr
dscottboggs: Generally, I prefer to make my tests as atomic and robust as possible, with little regard to the time it will take. However, in the case of these changes, you are right -- the slowdown is _significant_. I suspect the majority of that is from the time spent writing, deleting, and rewriting the file, but I'll do some checking and see if your idea will help significantly.

I was thinking earlier, perhaps instead of deleting the file every time, just rewrite it if the MD5 has does not match. This would be a read-only activity which could be cached by the Kernel. I will do some testing.

dscottboggs:
> For some reason this branch can't be built in the CI – it was stuck with a High GC allocation count warning.

probably due to the spec performance issue. The reason I had spun up a server in each spec was because the permissions_config.json has to be written before the server tries to start. Working on resolving those issues now.

dscottboggs:
Just checked it with the MD5 hash solution -- it increased the spec runtime from 26 to 32 seconds. :roll_eyes:

j8r:
~The CI is stuck on the build stage `shards build`.~ It's stuck at the end of `shards install`.
It's fine to let the `spec_helper` as is, the most inportant, as you said, is to have individual, atomic tests.

j8r:
After digging, good chances to be related to the [recent update of shards in Alpine to 0.9.0]( https://github.com/alpinelinux/aports/blob/master/community/shards/APKBUILD), done 2 days ago.

j8r:
Got it: https://github.com/crystal-lang/shards/issues/282#issuecomment-503152799

dscottboggs:
Interesting. Is there any reason not to merge in that case?

PR #20, opened 2019-06-19
Feature/error handling

### LINKED ISSUE #15: Error handling
Since the beginning of the project, I have been using the convention of returning an JSON object which has at its top-level some meaningful keys which our client can look for (I.E. `{"errors": ..., "data": ...}`). The "errors" key is always present -- in the case of no errors it is an empty list. The reason for it being a list is so that we can express to the user (or the author of a client library) a more meaningful error, by allowing the server to note a list of errors rather than a single one. 

However, the built-in error handling solutions in Crystal (raising exceptions) and in Kemal (exceptions, `halt`) are ideally designed to stop the execution of the context upon an error, without going on. My [initial idea](https://github.com/DFabric/dppm-rest-api/pull/13) was to store a list of errors on `HTTP::Server::Context`, but after looking through the overrides Kemal offers on `Context`, I thought of [this solution](https://github.com/DFabric/dppm-rest-api/pull/14).

Now that I think about it though, neither solution works at all. The result would look something like...
```json
{
  "data": { "some": "data" }
}
{
  "errors": ["an error", "Not Found"]
}
```
... which isn't even valid JSON.

Maybe if we check if `context.response` is empty (hasn't yet been written to) before writing to it in the [error handler](https://github.com/DFabric/dppm-rest-api/blob/eecbaf30649ae57fd14daf1ef73633fa4100a693/src/middlewares/error_handlers.cr). Idk, that still doesn't solve the issue of what to do inside `Actions.throw_error` and the `deny_access!` macro.

dscottboggs:
This will resolve #15


j8r:
`Exception#cause` can be used, like in https://github.com/DFabric/dppm/blob/master/src/logger.cr#L48, to implement https://jsonapi.org/examples/#error-objects-multiple-errors

dscottboggs:
Interesting... I'll get started on this feature.

Shouldn't [this](https://github.com/DFabric/dppm/blob/4e0eac5b81274fbccd9ed3f94afc67f34fa13a4a/src/logger.cr#L49) be `print_error ex.message`?

j8r:
In fact, [it's the same](https://github.com/crystal-lang/crystal/blob/master/src/exception.cr#L42).
Either way there would be an Union, message being nilable.

spec/spec_helper.cr
```diff
@@ -22,7 +34,7 @@ PERMISSION_FILE = Path[__DIR__, "permissions.json"]
 Spec.before_each do
   DppmRestApi.permissions_config = Fixtures.permissions_config
   DppmRestApi.permissions_config.write_to PERMISSION_FILE
-  DppmRestApi.run Socket::IPAddress::LOOPBACK, DPPM::Prefix.default_dppm_config.port, __DIR__
+  run_server unless already_running
```
j8r: Why? Also, a method isn't needed.

src/actions/app.cr
```diff
@@ -48,28 +44,27 @@ module DppmRestApi::Actions::App
       if key == "."
         dump_config context, app
       elsif config = app.get_config(key)
-        context.response.content_type = "application/json"
-        {"data" => config, "errors" => [] of Nil}.to_json context.response
+        {"data" => config}.to_json context.response
```
j8r: Can be a `NamedTuple` (more efficient than `Hash`)

src/config/jwt_compatible_hash.cr
```diff
@@ -0,0 +1 @@
+alias JWTCompatibleHash = Hash(String, String | Int32 | Bool?)
```
j8r: A file with a single alias? Why it isn't put in an already existing module, like `DppmRestApi`?

spec/spec_helper.cr
dscottboggs: > Why?

Because you only need to start the server once. Starting it multiple times may have unintended side effects. I thought it best to only start it once.

> a method isn't needed.

No, but it makes it neater and more readable, hence more maintainable in the future.

j8r: Ok. For the method, I don't think so, because we have to look back at another abstraction to see what is the actual code ran.
A method is useful if it's used multiple times.

But, why all of this instead of just having this call before `Spec.before_each`, as done first by me?

dscottboggs: so option A we can do

```crystal
already_running = false
Spec.before_each do
  DppmRestApi.permissions_config = Fixtures.permissions_config
  DppmRestApi.permissions_config.write_to PERMISSION_FILE
  DppmRestApi.run Socket::IPAddress::LOOPBACK, DPPM::Prefix.default_dppm_config.port, __DIR__ unless already_running
  already_running = true
end
```
or, alternatively --

```crystal
DppmRestApi.permissions_config = Fixtures.permissions_config
DppmRestApi.permissions_config.write_to PERMISSION_FILE
DppmRestApi.run Socket::IPAddress::LOOPBACK, DPPM::Prefix.default_dppm_config.port, __DIR__
Spec.before_each do
  DppmRestApi.permissions_config = Fixtures.permissions_config
  DppmRestApi.permissions_config.write_to PERMISSION_FILE
end
```
I didn't like the repetitiveness of the latter solution, that was my rationale for the first. The method was added because the last line of the before_each block is very long

dscottboggs: We could use the latter option, but with less repetition like

```crystal
def reset_config
  DppmRestApi.permissions_config = Fixtures.permissions_config
  DppmRestApi.permissions_config.write_to PERMISSION_FILE
end

reset_config
DppmRestApi.run Socket::IPAddress::LOOPBACK, DPPM::Prefix.default_dppm_config.port, __DIR__

Spec.before_each { reset_config }

File: src/config/jwt_compatible_hash.cr (hunk 3 above)
Comment (dscottboggs) [reply src/config/jwt_compatible_hash.cr]: I was trying to think of a place to put it where it fit with other stuff, and there just didn't seem to be any. It doesn't seem like a good idea to clutter the main file with stuff like this, and putting it anywhere else would make it hard to find. It doesn't do any harm being there. We could put it back in `src/utils.cr` but you seemed very much against the idea of having that at all.

File: spec/spec_helper.cr (hunk 1 above)
Comment (j8r) [reply spec/spec_helper.cr]: The latter option sounds good, one less conditional and boolean 👍

File: src/config/jwt_compatible_hash.cr (hunk 3 above)
Comment (j8r) [reply src/config/jwt_compatible_hash.cr]: You need it at `src/config/user.cr`, so it can be defined there.

Comment (dscottboggs) [reply src/config/jwt_compatible_hash.cr]: Ok, but if we find a reason to use that type outside of User then we should move it from there, but for now that looks fine.

File: spec/actions/app_spec.cr [hunk 4]
```diff
@@ -7,84 +7,140 @@ module DppmRestApi::Actions::App
       it "responds with 401 Unauthorized" do
         get fmt_route "/some-app/config/key"
         response.status_code.should eq 401
+        ErrorResponse.from_json(response.body)
```
j8r: All this lines can be a a helper method, like `assert_error_response(response)`

j8r: (inspired by https://github.com/crystal-lang/crystal/blob/master/spec/std/json/pull_parser_spec.cr#L88)

dscottboggs:
That is a good idea

spec/actions/app_spec.cr
dscottboggs: Implemented.

spec/spec_helper.cr
```diff
@@ -24,6 +24,14 @@ struct ErrorResponse
   property errors : Array(ErrorData)
 end
 
+def assert_unauthorized!(response : HTTP::Client::Response)
```
j8r: I don't get why there is an exclamation mark. There are usually used to mark a method as "dangerous", like `not_nil!`.

src/middlewares/error_handlers.cr
```diff
@@ -1,16 +1,44 @@
...
+        type:        error.class.to_s,
+        message:     error.message,
+        status_code: error.status_code,
+      }
+    else
+      # any generic exception
+      {
+        type:    error.class.to_s,
+        message: error.message,
+      }
```
j8r: Not a fan of returning a `NamedTuple`. I suggest a record, with `status_code` being nillable.
You can use the `ErrorData` from the `spec_helper` and put it there.

spec/spec_helper.cr
dscottboggs: This is dangerous method, it raises an exception to fail a test. Regardless, changed.

src/middlewares/error_handlers.cr
dscottboggs: Actually, I just moved the structs from spec_helper into that file, added a couple methods on them, and everything is neater. A nicer logging solution exposed itself in the process too :)

I don't like the `record` macro though, it seems silly, as it takes very little more to type out a `struct` with `property`s on it. It just seems unnecessary.

spec/spec_helper.cr
```diff
@@ -39,4 +39,4 @@ DppmRestApi.run Socket::IPAddress::LOOPBACK, DPPM::Prefix.default_dppm_config.po
 # Set all configs back to the expected values, in case they changed
 Spec.before_each { reset_config }
 # Clean up after ourselves
-Spec.after_each { File.delete PERMISSION_FILE if File.exists? PERMISSION_FILE }
+Spec.after_each { File.delete PERMISSION_FILE if File.exists? PERMISSION_FILE }
```
j8r: Please remove this commit

src/middlewares/error_handlers.cr
```diff
@@ -1,29 +1,59 @@
...
+  include JSON::Serializable
+  property type : String
+  property message : String?
+  property status_code : HTTP::Status?
+
+  def initialize(@type, @message, @status_code = nil)
   end
 
-  def response_data_for(error : Exception)
+  def self.from(error : Exception)
```
j8r: `from_exception` would be a better name, to differentiate with `from_json`

src/middlewares/error_handlers.cr
```diff
@@ -1,16 +1,74 @@
+require "http"
+require "json"
+
+struct ErrorData
+  include JSON::Serializable
+  property type : String
+  property message : String?
+  property status_code : HTTP::Status?
+
+  def initialize(@type, @message, @status_code = nil)
```
j8r:
```suggestion
  def initialize(@type : String, @message : String?, @status_code : HTTP::Status? = nil)
```

src/middlewares/error_handlers.cr
```diff
@@ -1,16 +1,74 @@
...
+      )
+    end
+  end
+end
+
+struct ErrorResponse
+  include JSON::Serializable
+  property errors : Array(ErrorData)
+
+  def initialize(@errors)
```
j8r:
```suggestion
  def initialize(@errors : Array(ErrorData))
```

src/middlewares/error_handlers.cr
```diff
@@ -1,16 +1,74 @@
...
+  include JSON::Serializable
+  property type : String
+  property message : String?
+  property status_code : HTTP::Status?
+
+  def initialize(@type, @message, @status_code = nil)
+  end
+
+  def self.from(error : Exception)
+    if error.responds_to? :status_code
```
j8r: I think it's possible to avoid this, by adding a method overload, more restrictive than `Exception`

src/middlewares/error_handlers.cr
dscottboggs: With JSON, one usually has an IO or file or something, so one calls `ErrorResponse.from_json(context.response)` or `ErrorResponse.from_json file`. `#from_exception` would stutter with a variable named `exception`, like in the block where it's called from:

```crystal
response_data = ErrorResponse.from_exception exception
```

I *really* don't like that. Can we compromise on a `self.new` overload?

src/middlewares/error_handlers.cr
dscottboggs: What is the point of this? A compiler error will still result if one calls `ErrorMessage.new` with a `Int32` as the first value. In fact, I had to make some changes to the types of this as I was developing it -- I originally had it like

```crystal
  property type : String
  property message : String
  property status_code : Int32?
```
and the compiler told me I couldn't call `ErrorMessage.new(String, String?, HTTP::Status?)`, as one would expect.

j8r: They will appear in the API docs. Maybe the compiler error is better, too.

src/middlewares/error_handlers.cr
j8r: Good idea yes

src/middlewares/error_handlers.cr
dscottboggs: updated

src/errors/http_status_errors.cr
```diff
@@ -0,0 +1,30 @@
...
+  class {{type_name.id}} < HTTPStatusError
+    STATUS_CODE = HTTP::Status::{{status_code}}
+
+    def status_code
+      STATUS_CODE
+    end
+
+    def self.status_code
+      STATUS_CODE
+    end
```
j8r: Why?

src/actions/pkg.cr
```diff
@@ -1,3 +1,5 @@
+require "../actions"
```
j8r: Not needed

src/actions/app.cr
```diff
@@ -1,5 +1,5 @@
 require "dppm"
-# require "../config/route"
+require "../actions"
```
j8r: Not needed

src/actions/src.cr
```diff
@@ -1,18 +1,20 @@
+require "../actions"
```
j8r: Not needed

src/actions/service.cr
```diff
@@ -1,3 +1,5 @@
+require "../actions"
```
j8r: Not needed

j8r (changes requested):

src/errors/http_status_errors.cr
dscottboggs: Why not?

src/actions/pkg.cr
dscottboggs: My editor hooks in with the compiler to alert me to issues with types or syntax while I'm working, and that feature is only effective if each file has the appropriate `require` statements. It doesn't hurt anything to be there.

j8r: This is a non sense, this file is required by a file (`actions.cr`), that is also required back?!
I don't know about your editor, but this isn't a strong enough argument against this anti-patterns.

Nobody wants to see this unneeded backward parent requirements everywhere.

src/errors/http_status_errors.cr
j8r: Because one can already use `STATUS_CODE`. Unless you have something against uppercase letters?

src/actions/pkg.cr
dscottboggs: :open_mouth: wow, if it's that important to you...

src/middlewares/error_handlers.cr
```diff
@@ -1,16 +1,76 @@
...
+  property message : String?
+  property status_code : HTTP::Status?
+
+  def initialize(@type : String,
+                 @message : String?,
+                 @status_code : HTTP::Status? = nil)
+  end
+
+  def self.new(error : Exception)
+    if error.is_a? DppmRestApi::HTTPStatusError
```
j8r, line 16: This doesn't work as a method overload (`def self.new(error : DppmRestApi::HTTPStatusError)`)?

dscottboggs:
Crap, I thought it didn't but it might have been due to a mistake elsewhere let me double check.

src/middlewares/error_handlers.cr
dscottboggs: Yeah, I just double-checked and indeed that does not work because `DppmRestApi::HTTPStatusError` inherits from `Exception`

src/actions/pkg.cr
j8r: Yes it is, I don't invent anything there, I'm looking at how other big projects like crystal, shards, lucky and other non-crystal projects.

That's a modularization principle: why here `pkg` will require `actions`, that will bring `src`, `app` etc?

Only import/require only what is needed – simpler to understand, avoid circle dependencies.

src/middlewares/error_handlers.cr
dscottboggs: I wish that worked... :slightly_frowning_face:

j8r (approved):

PR #21, opened 2019-06-21
Bug/route matching

dscottboggs:
~> `Duplicate trail found 'get/dppmrestapi/actions/app/:app_name/config/:key' (Radix::Tree::DuplicateError)`~

~We can't use this route for multiple endpoints. We can use `/app/:app_name/config/:key/{get,set,etc...}` or `/app/:app_name/config-{get,set,etc...}/:key`, but we can't use the same route multiple times with different verbs in Kemal. This is a limitation of using the Radix tree, and was why @Blacksmoke16 elected to not use a Radix tree for route matching in Athena.~

~Other than that hiccup,~ this PR resolves all routing issues and makes all formerly `pending` specs pass. :tada: :grin:

### DIFF
--- a/shard.lock
+++ b/shard.lock
@@ -42,7 +42,7 @@ shards:
 
   jwt:
     github: crystal-community/jwt
-    version: 0.4.0
+    version: 1.1.0
 
   kemal:
     github: kemalcr/kemal
@@ -63,6 +63,10 @@ shards:
   mysql:
     github: crystal-lang/crystal-mysql
     version: 0.6.0
+
+  openssl_ext:
+    github: stakach/openssl_ext
+    version: 1.0.0
 
   radix:
     github: luislavena/radix
--- a/shard.lock
+++ b/shard.lock
@@ -22,7 +22,7 @@ shards:
 
   dppm:
     github: DFabric/dppm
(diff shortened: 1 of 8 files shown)


j8r:
That's problematic. We have to find a way to use several HTTP verbs on the same routes, else there is no point of having them.
We have to keep the API as clean as consistent as possible.

Blacksmoke16:
Athena also gets around this by internally prepending the method to the beginning of the string. I.e. `GET /foo` is stored as `/GET/foo`.  So that when a request comes in I can tie the method to the correct route, even if they're the same route; of course you would still run into this if you had two GET requests with the same path, and neither used constraints.

Something like:
```
route = @routes.find '/' + ctx.request.method + ctx.request.path
```

Internally it looks like that is what Kemal does as well.
https://github.com/kemalcr/kemal/blob/master/src/kemal/route_handler.cr#L62-L64

Are you sure you just don't have a copy paste error somewhere?

dscottboggs:
@Blacksmoke16 was right, there was an extra `get` block.

j8r (approved):
Needs a `shards update`, then ready to go

PR #22, opened 2019-06-21
Refactor modularization

j8r:
This opens the way for more various, independent module tests.
We can now test different parts of `Config`, without needing anything to set up.


src/actions.cr
```diff
@@ -40,4 +45,6 @@ module DppmRestApi::Actions
   def self.deny_access!(to context)
     raise Unauthorized.new context
   end
+
+  protected class_property? has_access : Proc(HTTP::Server::Context, Access, Bool) = ->(context : HTTP::Server::Context, access : Access) { false }
```
dscottboggs: I like the idea here -- allowing overriding the block to change how access is granted (for testing I assume)... but I think the implementation could use some improvement. Instead of calling `#call` on the proc directly, I think the property should be called `access_filter` or something to that effect, has_access? can be like
```crystal
def has_access?(context, access)
  access_filter.call context, access
end
```
Secondly, I think the property should be on the authentication middleware, so we should move `DppmRestApi::Actions.auth_handler` to `DppmRestApi::Authentication.handler`. Then there can be an `Authentication.has_access?` next to it.

src/cli.cr
```diff
@@ -280,7 +280,7 @@ module DppmRestApi::CLI
     DppmRestApi.run host, port, data_dir
   end
 
-  def add_user(name, groups, data_dir, output_file, **args) : Void
+  def add_user(name, groups, data_dir, output_file, **args) : Nil
```
dscottboggs, line 283: Why?

src/config.cr
```diff
@@ -11,7 +10,8 @@ struct DppmRestApi::Config
   property users : Array(User)
 
   # :nodoc:
-  def initialize(@groups, @users)
+  def initialize(@groups : Array(Group), @users : Array(User))
+    User.all_groups = @groups
```
dscottboggs: This would not get called when deserializing from JSON.

src/config/user.cr
```diff
@@ -5,21 +5,19 @@ struct DppmRestApi::Config::User
   API_KEY_SIZE = 63_u8
   include JSON::Serializable
   property api_key_hash : Scrypt::Password
-  setter groups : Set(Int32)
+  getter group_ids : Set(Int32)
+  property name : String
+  # :nodoc:
+  class_property all_groups : Array(Group) = Array(Group).new
```
dscottboggs: What's the point of having a second copy of this?

src/dppm_rest_api.cr
```diff
@@ -18,6 +15,20 @@ module DppmRestApi
     end
 
     Kemal.config.add_handler Actions.auth_handler
+
+    Actions.has_access = ->(context : HTTP::Server::Context, permission : Access) {
```
dscottboggs: If we're going to make this change, we should have tests specifically for this proc, and in `spec_helper`, we should do `Actions.has_access = ->(context : HTTP::Server::Context, permission : Access) { true }` and expect a status code of 200.

dscottboggs (changes requested):
I really like this change. Really good idea.

src/cli.cr
j8r: See https://crystal-lang.org/reference/syntax_and_semantics/return_types.html.

> Using `Void` is the same, but `Nil` is more idiomatic: `Void` is preferred in C bindings.

src/config/user.cr
j8r: That's not a copy, array is a class, so passed as a reference.
How do you access to all groups from an `User` instance then (without depending to `DppmRestApi`)?

src/cli.cr
dscottboggs: Fair enough.

src/config/user.cr
dscottboggs: Oh, ok, you're right. Since it's just a reference it's no big deal. I still don't know how to do this when deserializing from JSON though...

j8r: I've changed the approach by using a `GroupView`

src/dppm_rest_api.cr
j8r: How to do this? I've tried something in `actions_spec.cr`

src/dppm_rest_api.cr
```diff
@@ -29,7 +29,23 @@ module DppmRestApi
       false
     }
 
-    initialize_error_handlers
+    # Add authentification handler
```
dscottboggs, line 39: I think I prefer to keep the middleware as a separate file at least. This section will only grow over time. Is there a reason we should prefer this form of keeping fewer files?

src/actions.cr
```diff
@@ -42,5 +41,9 @@ module DppmRestApi::Actions
     end
   end
 
-  protected class_property? has_access : Proc(HTTP::Server::Context, Access, Bool) = ->(context : HTTP::Server::Context, access : Access) { false }
+  def self.has_access?(context : HTTP::Server::Context, access : Access) : Bool
+    access_filter.call context, access
+  end
```
dscottboggs, line 46: :+1: :100:

dscottboggs (commented):
This is really great. A couple things I would add:

```crystal
# In src/dppm_rest_api.cr
def default_acces_filter(context : HTTP::Server::Context, permission : Access) : Nil
  # (what you had)
end
def self.run(*args )
  run *args, &->default_access_filter(HTTP::Server::Context, Access)
end
def self.run(host : String, port : Int32, data_dir : String, &block : HTTP::Server::Context, Access -> Nil)
  # (what was already there, except)
  Actions.access_filter = ->block
end
```
Then, in `spec_helper`, we could do `DppmRestApi.run(host, port, __DIR__) { true }` to allow all requests for testing purposes.

src/dppm_rest_api.cr
j8r: Better having all the code at the same place instead of multiple arbitrary small files. This file is about 50 lines, which is totally fine.

We can move this away in the future if there are good reasons, for now I don't see any.

j8r:
Done. Allow all requests in tests?

spec/spec_helper.cr
```diff
@@ -25,6 +25,15 @@ macro assert_unauthorized(response)
   fail "expected error response not found" unless found
 end
 
+# Disable all authentication checks by making `DppmRestApi::Actions.has_access?`
+# always return true, then set the authentication back to the default system
+# after yielding to the block.
+def without_authentication!
```
dscottboggs, line 31: Is this ok? If it is, I'm ready to merge this when you are.

dscottboggs (approved):

spec/spec_helper.cr
j8r: Sounds good. Any idea to test the filter in `actions_spec.cr`?
We can fix it in a further PR if we don't find a solutiom yet

dscottboggs: Well, the access_filter has to take several steps, each of which are already tested more or less. I agree that an integration/regression test on this would be due, but I think it would be better to set that up in a PR dedicated to integration and regression tests once the API is a little more fleshed out.

If you'd prefer, I could set up a test for this method before merging it.

j8r: I agree. I think we are good to merge, this will be simpler.

PR #25, opened 2019-07-04
Feature/user actions

### LINKED ISSUE #9: User/Group Modification REST endpoints
To allow editing user permissions from the web UI or other end-user application.

- [ ] _POST_ `/user` 
  **JSON body**: user's name and group membership.
  **Accepted Keys**: `[ "name", "groups" ]`
  **Returns**: API key.
- [ ] _PUT_ `/users`
  **JSON body**: selectors and modifiers for users.
  **Accepted Keys**: `[ "match_name", "match_groups", "api_key", "new_name", "add_groups", "remove_groups" ]`
- [ ] _PUT_ `/user` edit the current user
  **JSON Body**: modifiers for the user
  **Accepted Keys**: `[ "new_name", "add_groups", "remove_groups" ]`
- [ ] _DELETE_ `/users`
  **JSON body**: of selectors for users
  **Accepted Keys**: `[ "match_name", "match_groups", "api_key" ]`
- [ ] _GET_ `/user`
  **Returns**: information about the currently signed in user
- [ ] _GET_ `/users`
  **JSON body**: selectors
  **Returns**: JSON-formatted list of user information
- [ ] _PUT_ `/user/rekey`
  **Returns**: A new API key for the currently signed-in user.
- [ ] _PUT_ `/users/rekey`
  **JSON body**: selectors
  **Returns**: the new user's API key.
  **Error**: if more than one user was selected.
- [ ] _POST_ `/groups`
  **JSON body**: the group
- [ ] _PUT_ `/groups/:id/route`
  **JSON body**: a new route to add to the group
- [ ] _PUT_ `/groups/:id/param`
  **JSON body**: a mapping of a path to a list of globs to add to those available on that path
- [ ] _DELETE_ `/groups/:id/param`
  **JSON body**: a group mapping a path to a list of globs to remove
- [ ] _DELETE_ `/groups/:id`
  Deletes the group

dscottboggs:
I'm beginning work on the user/group routes. Once I have the `get` routes down we should be able to start work on the web interface as well.

This will resolve #9.


src/actions.cr
```diff
@@ -7,14 +7,14 @@ module DppmRestApi::Actions
   extend self
   include RouteHelpers
 
+  ROUTE_MODULES    = [Pkg, App, Service, Src, User, Users, Groups]
```
j8r: This constant doesn't seem to be useful, as-is

src/actions.cr
```diff
@@ -7,14 +7,14 @@ module DppmRestApi::Actions
...
   DEFAULT_DATA_DIR = "./data/"
   API_DOCUMENT     = DEFAULT_DATA_DIR + "api-options.json"
   alias ConfigKeyError = DPPM::Prefix::Base::ConfigKeyError
 
-  include Pkg
-  include App
-  include Service
-  include Src
+  {% for route_module in [Pkg, App, Service, Src, User, Users, Groups] %}
```
j8r: It was perfectly good before, we have to write the module names anyway

src/actions.cr
dscottboggs: Yes it was supposed to be for the other thing.

src/actions.cr
dscottboggs: TBH I forgot I left this here when I pushed this morning... I should've known you wouldn't like it :p

I *always* do this sort of thing in personal projects. Have to type the same thing more than 3 or 4 times? Nope, that goes in a macro. Haha. I'll change it back when I go to make some changes on this next.

j8r: I agree on the concept, but here in thia very case thats only `include` to add – doesn't worth the cost of macros IMO. There are even almost no characters saved, if it was the main point (at least for now)

j8r: I agree on the concept, but here in thia's very case thats only `include` to add – doesn't worth the cost of macros IMO. There are even almost no characters saved, if it was the main point (at least for now)

j8r: I agree on the concept, but here in this very case thats only `include` to add – doesn't worth the cost of macros IMO. There are even almost no characters saved, if it was the main point (at least for now). There are still hardcoded somehow.

src/utils.cr
```diff
@@ -0,0 +1,103 @@
+require "./dppm_rest_api"
```
j8r: What is this file, why requiring the whole application?

src/utils.cr
```diff
@@ -0,0 +1,103 @@
...
+
+# A helper method for the select_users method -- splits a comma-separated
+# list of numbers in a string into a `Set` of `Int32` values.
+def split_numbers(number_list : String) : Set(Int32)
```
j8r: Should be namespaced in CLI, there are only used there.

dscottboggs: That's not the case anymore. It is used by the `/{user, users,groups}` routes now. That is why these functions have been moved to this file.

To be honest I expected you to be dissatisfied with this solution and meant to ask you for suggestions of how to resolve this issue. Specifically, the `/user` routes need `selected_users`

j8r: The CLI depends of the whole application, but not actions. They can be put in actions where they are used, or a new module.
Having them globally won't help the understanding and the API docs.

dscottboggs:
I'll have to look into that merge conflict later. :frowning_face:

j8r (changes requested):
`def_clone` can be dropped

src/actions/user.cr
```diff
@@ -0,0 +1,87 @@
...
+    end
+    relative_get do |context|
+      raise Unauthorized.new context unless Actions.has_access? context, Access::Read
+      build_json context.response do |response|
+        response.field "users" do
+          selected_users_from_query.to_json response
+        end
+      end
+    end
+    get "/me" do |context|
```
j8r: no `relative_get`?
A more generic `/self` would be preferable in my opinion, that can be also used in other endpoints.

dscottboggs:
> `def_clone` can be dropped

where?

src/actions/user.cr
dscottboggs: Sure, I'm not super concerned about what API endpoint is used for what. I chose `/me` here because that's the default on a generated Lucky app. You think it should be `relative_get "/self"`?

src/config.cr
```diff
@@ -53,6 +61,8 @@ struct DppmRestApi::Config
       to_json builder
     end
   end
+
+  def_clone
```
j8r: To remove

src/config/group.cr
```diff
@@ -39,4 +39,5 @@ struct DppmRestApi::Config::Group
       Access::All,
       {"namespace" => ["default-namespace"]}),
   }
+  def_clone
```
j8r: To remove

src/config/route.cr
```diff
@@ -1,6 +1,7 @@
 # Matches a request to a series of globs loaded from a config file.
 struct DppmRestApi::Config::Route
   include JSON::Serializable
+  def_clone
```
j8r: To remove

src/config/user.cr
```diff
@@ -70,9 +69,30 @@ struct DppmRestApi::Config::User
...
+      {% end %}
+    end
+  end
+
+  # :ditto:
+  def to_json(builder : JSON::Builder, *, except : Symbol)
+    to_json builder, except: StaticArray[except]
+  end
+
+  def_clone
```
j8r: To remove

src/ext/path.cr
```diff
@@ -0,0 +1,11 @@
+struct Path
+  def to_json(builder : JSON::Builder) : Nil
+    builder.string self.to_s
+  end
+
+  def self.new(json : JSON::PullParser)
+    new json.read_string
+  end
+
+  def_clone
```
j8r: To remove

src/ext/scrypt_password.cr
```diff
@@ -6,4 +6,5 @@ class Scrypt::Password
   end
 
   delegate :to_json, to: @raw_hash
+  def_clone
```
j8r: To remove

src/ext/uuid.cr
```diff
@@ -0,0 +1,6 @@
+require "uuid"
+require "uuid/json"
+
+struct ::UUID
+  def_clone
+end
```
j8r: To remove

src/config/user.cr
```diff
@@ -1,6 +1,5 @@
 require "json"
-require "uuid"
-require "uuid/json"
+require "../ext/uuid"
```
j8r: To remove

j8r (dismissed):

src/actions/user.cr
j8r: Not sure, `relative_get "/me"` may be good too, since there is no `/group` endpoint?

dscottboggs:
Regarding `def_clone`, this

```crystal
module Fixtures
  ...
  def reset_config
    DppmRestApi.permissions_config = PERMISSIONS_CONFIG.clone
  ...
```

was still in this branch. Just rebased to take care of that.

src/actions/user.cr
dscottboggs: Group access and modification routes will be a part of another PR.

dscottboggs:
Also sorry idk why I said "Where?" last night, what I meant to say was something along the lines of "don't we need that?" but it's solved now anyway so never mind haha.

src/ext/uuid.cr
j8r: Still not removed :wink:

spec/spec_helper.cr
```diff
@@ -25,6 +25,19 @@ macro assert_unauthorized(response)
...
 end
 
+def assert_no_error(in response : HTTP::Client::Response) : Nil
+  return if response.status_code == 200
+  response_errors = String.build do |str|
+    str << "Failure! Reponse status code was #{response.status_code}\n"
+    DppmRestApi::Actions::ErrorResponse.from_json(response.body).errors.each do |error|
+      str << '\t' << error.type << ": " << error.message
+      str << "\n\t resulting in status_code code " << error.status_code if error.status_code
+      str << '\n' << '\n'
```
j8r:
```suggestion
      str << "\n\n"
```

spec/actions/user_spec.cr
```diff
@@ -0,0 +1,119 @@
...
+          end
+        end
+      end
+    end
+
+    include JSON::Serializable
+    getter data : Data
+
+    def should_contain_user_named(name : String)
+      user = data.users.find { |u| u.name == name }
```
j8r:
```suggestion
      user = data.users.find &.name.== name
```

spec/actions/user_spec.cr
```diff
@@ -0,0 +1,119 @@
...
+            when "SuccessfullyAddedUser" then user = UserAddResponse.new json
+            else
+              fail "received unexpected response key #{json_key}"
+            end
+          end
+        end
+        fail "response did not contain AccessKey" if key.nil?
+        fail "response did not contain SuccessfullyAddedUser" if user.nil?
+        user.name.should eq "Mock user"
+        if config_user = DppmRestApi.permissions_config.users.find { |u| u.name == user.name }
```
j8r:
```suggestion
        if config_user = DppmRestApi.permissions_config.users.find &.name.== user.name
```

j8r (commented):
Some pedantic changes

spec/actions/user_spec.cr
dscottboggs: Will this work? I thought it wouldn't work unless you use parens like `&.name.==(name)`. Regardless, I think the full block syntax is less confusing when calling _operators_ on block data.

src/actions/user.cr
j8r: Absolutely. Remains a little test for this endpoint.

spec/actions/user_spec.cr
j8r: It works.
`u` doesn't provide any information. We already know this are users, because of `data.users`. I don't think it's confusing, the syntax is quite common.

src/ext/uuid.cr
dscottboggs: Weird...

j8r (approved):

PR #27, opened 2019-07-06
Feature/src actions

dscottboggs:


dscottboggs:
How can I test if a `Dppm::Prefix::Src` is a lib or app?

j8r:
For what purpose? You can use `Src#pkg_file.type.lib?`

dscottboggs:
> For what purpose?

https://github.com/DFabric/dppm-rest-api/blob/4eecfeb31a8ec53b88b1df88daf17f609044f080/src/actions/src.cr#L11

dscottboggs:
I'm in the process of implementing [this route](https://github.com/DFabric/dppm-rest-api/blob/4eecfeb31a8ec53b88b1df88daf17f609044f080/src/actions/src.cr#L11). I'm thinking, perhaps `#type` is one of many attributes which should be optionally queryable on the  `/api/actions/src` root route, and perhaps one should be able to filter by those values as well.

My point is that perhaps having a separate route for `#type` specifically doesn't make a ton of sense. Instead, we should focus on a macro implementation which DRYly iterates over all of the `getter`s in `DPPM::Prefix::PkgFile`  for filtering/querying all values.

j8r:
Note that there is the source name, then the packages source inside it.
I think the `/pkg`, `/src` and `/app` should list current package group, source names, and application namespaces.
Then, there will be `/src/:source/:source_package`

dscottboggs:
So you agree that `/src/:type` is a silly route to have?

j8r:
That's a good idea to use a macro, but I don't think it's needed.
`DPPM::Prefix::PkgFile.new(type : String)` can be used, then compare the type with the source packages.
I wonder how to list only applications, without having to have to list all types (`html`, `http`, etc) but `lib`.
We could introduce `app`.

dscottboggs:
I was actually considering having an additional `app` route, which would just check if the `src` is not a `lib`, but again, I feel like this is more appropriate to be done in optional query parameters under a single route.

src/actions/src.cr
```diff
@@ -1,19 +1,96 @@
...
+    def to_json(builder : JSON::Builder, only : Enumerable(String))
+      if only.empty?
+        to_json builder
+        return
+      end
+      {% for ivar in @type.instance_vars %}
+        if only.includes? {{ivar.name.stringify}}
+          builder.field({{ivar.name.stringify}}) { @{{ivar.id}}.to_json builder }
+        end
+      {% end %}
```
j8r: You can use `each_ivar`

src/actions/src.cr
```diff
@@ -1,19 +1,96 @@
...
+    def each_ivar
+      {% for ivar in @type.instance_vars %}
+        yield {{ ivar.stringify }}, @{{ivar.id}}
+      {% end %}
+    end
+
+    # Build a JSON response of the config keys mapped to their values, but only
+    # if they're specified in the *only* list. If *only* is empty, all values
+    # are forwarded
+    def to_json(builder : JSON::Builder, only : Enumerable(String))
```
j8r: `only` isn't self-explanatory enough.
I suggest to use a block for filtering, like `Enumerable#select`

dscottboggs: I don't like the idea of using a block. It would either be less clear, in the case of a syntax like...
```crystal
src.pkg_file.to_json response do |key|
```
...because it's not super clear what it is we're *do*ing with _key_. In the case of a syntax like...
```crystal
src.pkg_file.select { |key| ... }.to_json response
```
...it would require building of an intermediate data structure, and even if it's just `Iterator.of`, that has some significant overhead associated with it over the simple `each_ivar` macro.

That said I agree the wording could be more clear. I thought I had seen `only`, and [`except`](https://github.com/DFabric/dppm-rest-api/blob/2defa42b63eaa5ee6e6c4a6dcb75b0e3e0400106/src/config/user.cr#L80) somewhere, but I can't recall where, and there's no reason _seeing it before_ makes something gospel haha. What about `expose` and `redact`?

j8r: Replacing `only` with `keys` would be enough to clarify

src/actions/src.cr
```diff
@@ -1,19 +1,96 @@
+require "dppm/prefix/src"
+
 module DppmRestApi::Actions::Src
   extend self
   include RouteHelpers
 
+  struct MultiSet(K, V)
```
j8r: Why creating a whole new type only used in `filters_and_return_keys`?

src/actions/src.cr
```diff
@@ -1,19 +1,96 @@
...
+        @data[k] &= v.to_set
+      else
+        @data[k] = Set(V).new v
+      end
+    end
+
+    forward_missing_to @data
+  end
+
+  struct ::DPPM::Prefix::PkgFile
```
j8r: This isn't only needed for `src`, but also `pkg` and `app`.
Creating a package copy the source one, and creating an app may copy a package if self-contained.

src/actions/src.cr
dscottboggs: Because it isn't an existing stdlib type, but this is the data structure which best encapsulates the data.

j8r: Ok, but this means you have to add tests for it.
The logic implemented by it could be in the method, what advantage does it add?
One disadvantage is to have the logic in two places.

j8r: Is it possible to group all filter related stuff inside a struct, which have both this logic, `filters_and_return_keys` and `filters`?

src/actions/src.cr
```diff
@@ -1,26 +1,113 @@
+require "../ext/path"
```
j8r: It will likely be needed for other actions as well. Can be moved to `actions.cr`
It will also avoid a backward reference.

dscottboggs:
Please let me know what you think of the revised structure of this PR.

src/actions/errors/invalid_filter.cr
```diff
@@ -0,0 +1,12 @@
+module DppmRestApi::Actions
+  class InvalidFilter < BadRequest
+    MESSAGE = "A filter MUST contain at least two equals-separated values, \
+    like filter=somekey=value. You can specify more than one option, like \
+    filter=somekey=value1=value2=value3, but there must be at least a key \
+    and a value."
+
+    def initialize(context, cause : Exception? = nil)
+      super context, MESSAGE, cause
```
j8r: The message can be directly put there, instead of using a constant.

src/actions/route_helpers/filters.cr
```diff
@@ -0,0 +1,52 @@
+# A series of filters, which may be parsed from the current context. They
+# currently allow filtering the #pkg_file attributes of DPPM::Prefix's Pkg,
+# App, and Src.
+class DppmRestApi::Actions::RouteHelpers::Filters < Hash(String, Set(String))
```
j8r: This is considered a bad practise, see https://github.com/crystal-lang/shards/pull/230.
Use composition instead.
Just having an internal ivar Hash will perfectly do the job.

src/ext/dppm_prefix_pkg_file.cr
```diff
@@ -0,0 +1,49 @@
...
+
+struct ::DPPM::Prefix::PkgFile
+  # Yield each instance variable to the block
+  def each_ivar
+    {% for ivar in @type.instance_vars %}
+        yield {{ ivar.stringify }}, @{{ivar.name}}
+      {% end %}
+  end
+
+  NOT_SERIALIZABLE = {"any", "config", "tags", "version", "tasks"}
```
j8r: This should not be a public constant, but a macro variable.

src/actions/route_helpers/filters.cr
```diff
@@ -0,0 +1,52 @@
...
+    keep = true
+    pkg_file.each_ivar do |key, value|
+      if filter_set = self[key]?
+        keep = false unless filter_set.includes? value
+      end
+    end
+    keep
+  end
+
+  {% for type in {:src, :app, :pkg} %}
```
j8r:
```suggestion
  {% for type in %w(src app pkg) %}
```

src/actions/route_helpers.cr
```diff
@@ -78,3 +78,5 @@ module DppmRestApi::Actions::RouteHelpers
     ).tap &.ensure_pkg_dir
   end
 end
+
+require "./route_helpers/**"
```
j8r:
```suggestion
require "./route_helpers/*"
```
Not a fan of `**`, which import every sub-files.
Better to only use `*`, this limit issues related to file require order.

src/actions/route_helpers/filters.cr
```diff
@@ -0,0 +1,52 @@
...
+  # build the JSON list of data from the package files of all {{type.id}}s
+  # that pass the filters.
+  def {{type.id}}s_json(builder : JSON::Builder, keys : Enumerable(String))
+    builder.array do
+      Actions.prefix.each_{{type.id}} do |x|
```
j8r:
```suggestion
      Actions.prefix.each_{{type.id}} do |{{type.id}}|
```

src/actions/route_helpers/filters.cr
```diff
@@ -0,0 +1,52 @@
...
+        x.pkg_file.to_json builder, keys if filters_allow? x.pkg_file
```
j8r:
```suggestion
        {{type.id}}.pkg_file.to_json builder, keys if filters_allow? {{type.id}}.pkg_file
```

src/actions/route_helpers/filters.cr
dscottboggs: I thought of doing this, but was unsure which made more sense. One makes as much sense to me as the other :man_shrugging:

src/actions/route_helpers/filters.cr
dscottboggs: > Inheriting from stdlib classes is considered a bad practice

What??? This seems like some incredibly backwards thinking, especially for an OOP language. Using composition is a minor refactor, so I'll do it, but this makes very little sense to me.

j8r: In any way this shouldn't be a Hash, this is an implementation detail.
In my opinion the Filter type should be a type of its own.

We want the use to use it as a Filter, never as another Hash.
We will also be free to change it to any other type if optimization is needed.

j8r: In any way this shouldn't be a Hash, this is an implementation detail.
In my opinion the Filter type should be a type of its own.

We want the user to use it as a Filter, never as another Hash.
We will also be free to change it to any other type if optimization is needed.

dscottboggs: Fair points.

src/ext/dppm_prefix_pkg_file.cr
```diff
@@ -0,0 +1,47 @@
+require "dppm/prefix"
+
+struct ::DPPM::Prefix::PkgFile
+  # Yield each instance variable to the block
+  def each_ivar
+    {% for ivar in @type.instance_vars %}
+        yield {{ ivar.stringify }}, @{{ivar.name}}
+      {% end %}
```
j8r: Little indent mismatch :wink:

j8r (approved):
Good job @dscottboggs :+1:

PR #28, opened 2019-07-07
Use Fixtures instances for isolated specs

j8r:


spec/spec_helper.cr
```diff
@@ -38,16 +38,20 @@ DPPM::Log.output = DPPM::Log.error = File.open File::NULL, "w"
...
-Fixtures.reset_config
+Fixtures.new.reset_config
 # Run the server
 DppmRestApi.run Socket::IPAddress::LOOPBACK,
   DPPM::Prefix.default_dppm_config.port,
   Fixtures::DIR,
   DPPM::Prefix.new(Fixtures::PREFIX_PATH.to_s).tap &.create
+
+# Having a pool of pre-allocated Fixtures cut to a least 2 times the total execution time
+fixtures_pool = Array(Fixtures).new 999, Fixtures.new
```
dscottboggs:
```suggestion
fixtures_pool = StaticArray(Fixtures, 999).new Fixtures.new
```

dscottboggs (commented):
:+1:  Nice

spec/spec_helper.cr
j8r: Was advantage does it bring? As is, `shift` will no longer work.
It could bring some memory and performance improvement, but I doubt there will be significant, especially for specs.

j8r:
I am wondering what test can we add to ensure this is working as expected?

spec/spec_helper.cr
dscottboggs: It reduces memory consumption. However, if there is a disadvantage (like no shift) then nevermind.

dscottboggs:
> I am wondering what test can we add to ensure this is working as expected?

The way that I found that we *weren't* actually resetting fixtures was by taking a reference of it (`...permisssions_config = ...`), then mutating something under that reference (`permissions_config.users << User.new...`), and checking to see if the underlying value has changed after calling `Fixtures.reset_config`

spec/spec_helper.cr
j8r: After a test, I've realized this approach doesn't work.
Back to the good old `Fixtures.new.reset_config`

dscottboggs (approved):

PR #29, opened 2019-07-07
Use source name route

j8r:


src/actions/pkg.cr
```diff
@@ -48,22 +38,31 @@ module DppmRestApi::Actions::Pkg
...
+
   # Clean unused built packages
-  relative_delete "/clean" do |context|
+  relative_delete "/:source_name/clean" do |context|
     if Actions.has_access? context, Access::Delete
-      clean_unused_packages(context).to_json context.response
-      context.response.flush
-      next context
+      prefix = get_prefix_with_source_name context
+      result = prefix.clean_unused_packages(confirmation: false) { }
```
dscottboggs, line 46: Has the implementation changed to return an empty set instead of `nil`?

spec/actions/pkg_spec.cr
```diff
@@ -61,36 +56,36 @@ module DppmRestApi::Actions::Pkg
   describe DppmRestApi::Actions::Pkg do
     describe "list built packages" do
       it "responds with 401 Forbidden" do
-        get fmt_route nil
+        get fmt_route '/' + DPPM::Prefix.default_source_name
```
dscottboggs, line 59: What is this?

src/actions/route_helpers.cr
```diff
@@ -67,4 +67,16 @@ module DppmRestApi::Actions::RouteHelpers
...
+
+  # Get a prefix from a context, which has a source name.
+  def get_prefix_with_source_name(context : HTTP::Server::Context) : DPPM::Prefix
+    prefix = DPPM::Prefix.new(
+      path: Actions.prefix.path.to_s,
+      group: Actions.prefix.group,
+      source_name: URI.unescape(context.params.url["source_name"]),
+      source_path: Actions.prefix.source_path
+    )
+    prefix.ensure_pkg_dir
```
dscottboggs: Since everything returns in crystal, I try to make methods like these chainable (end with returning `self`) when it makes sense (like here) or otherwise mark them as returning nil so that one doesn't unintentionally try to use the value of it. Just a suggestion.

src/actions/src.cr
```diff
@@ -1,13 +1,29 @@
 module DppmRestApi::Actions::Src
   extend self
   include RouteHelpers
-  relative_get "" do |context|
+
+  # Lists available sources.
```
dscottboggs, line 5: This renders #27 redundant :thinking:

src/actions/src.cr
```diff
@@ -1,13 +1,29 @@
...
+          end
+        end
+      end
       next context
     end
     raise Unauthorized.new context
   end
+
   # List all available source packages, of either the *lib* or *app* type.
   relative_get "/:type" do |context|
```
dscottboggs: Can we get rid of this route?

spec/actions/pkg_spec.cr
j8r: We get all source packages from the default source.

src/actions/route_helpers.cr
j8r: By using `.tap &.`?

src/actions/pkg.cr
j8r: Yes, I looked into this and indeed always returning a `Set` instead of an union with `Nil` sounds better.

spec/actions/pkg_spec.cr
j8r: (Not implemented yet)

src/actions/src.cr
j8r: Only very partially. I have only done one route for listing available sources here.
We will need source package information too, and listing them.

src/actions/route_helpers.cr
dscottboggs: no, like

```
  # Ensure the presence of the `pkg` directory.
  def ensure_pkg_dir
    Dir.mkdir(@pkg.to_s) if !File.exists? @pkg.to_s
    self
  end
```

spec/actions/pkg_spec.cr
j8r: We get all built packages from a given source, here the default one.

src/actions/route_helpers.cr
j8r: Why we will return `self`? `Nil` is fine, that's a file system operation. Maybe returning `@pkg` but I don't see the use case.

dscottboggs: I was suggesting to return self because of the use case. Statements like these, which are a guard against an error case almost always are immediately followed by doing something on or returning the value which was just checked by them. Here we do `prefix.ensure_pkg_dir; prefix`, and that is likely to be the pattern used by any code which calls `#ensure_pkg_dir`. Like with the stdlib's `not_nil!`?

Like I said it's just a suggestion, something I picked up from other Crystal projects that I've found to be helpful.

j8r: I have no strong opinion on this, as long as the return type is documented.
I've used `tap` here to shorten the expression. I may do the change when there will be an opportunity to benefit of it.

dscottboggs: CI is failing due to this

dscottboggs (approved):

PR #30, opened 2019-07-10
Add UUID to user

dscottboggs:
I think this will be much better than trying to identify a user by their API key or fuzzy-matching users for actions which mutate them. Additionally, I propose that we remove the fuzzy-matching from all of the routes and CLI actions except for one, which returns the matching users so you can get the UUID to reference in the other routes/CLI actions.

### DIFF
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -36,7 +36,9 @@ module DppmRestApi::CLI
             info:      "Update or delete users",
             inherit:   \%w(data_dir),
             variables: {
-              user_id: "a users's UUID"
+              user_id: {
+                info: "a users's UUID"
+              }
             },
             commands: {
               add: {
@@ -314,8 +316,7 @@ module DppmRestApi::CLI
     output_io.puts "API key for user named '#{name}'"
     output_io.puts api_key
     if using_stdout
-      print "ok? (Y/n) "
-      raise "cancelled" if gets(chomp: true).try &.downcase.starts_with?('n')
+      DPPM::CLI.confirm_prompt { raise "cancelled" }
     else
       output_io.close
     end
@@ -323,39 +324,35 @@ module DppmRestApi::CLI
     current_config.write_to permissions_file
   end
 
-  def edit_users(match_name, match_groups, api_key, new_name, add_groups, remove_groups, data_dir)
-    required data_dir
+  def edit_users(user_id, new_name, add_groups, remove_groups, data_dir)
+    required data_dir, user_id
     permissions_file = Path[data_dir, "permissions.json"]
     current_config = File.open permissions_file do |file|
       DppmRestApi::Config.from_json file
     end
-    selected_users match_name, match_groups, api_key, from: current_config.users do |user|
-      modified = false
-      # edit the users that have made it through all the filters, based on the
-      # non-nil arguments given by the user
-      new_name.try do |specified_new_name|
-        modified = true
-        user.name = specified_new_name
-      end
-      add_groups.try do |groups_to_add|
-        split_numbers groups_to_add do |id|
-          modified = true
-          user.join_group id
-        end
-      end
-      remove_groups.try do |groups_to_remove|
-        split_numbers groups_to_remove do |id|
-          modified = true
-          user.leave_group id
-        end
-      end
-      modified ? user : nil
+    user_index = current_config.users.index { |usr| usr.id == user_id }
+    raise "no user found with id #{user_id}" if user_index.nil?
+    user = current_config.users[user_index]
+    # edit the user based on the non-nil arguments given by the user
+    new_name.try do |specified_new_name|
+      user.name = specified_new_name
     end
+    add_groups.try do |groups_to_add|
+      split_numbers groups_to_add do |id|
+        user.join_group id
+      end
+    end
+    remove_groups.try do |groups_to_remove|
+      split_numbers groups_to_remove do |id|
+        user.leave_group id
+      end
+    end
+    current_config.users[user_index] = user
     current_config.write_to permissions_file
   end
 
-  def rekey_users(match_name, match_groups, api_key, data_dir, output_file)
-    required data_dir
+  def rekey_users(user_id, data_dir, output_file)
+    required data_dir, user_id
     permissions_file = Path[data_dir, "permissions.json"]
     current_config = File.open permissions_file do |file|
       DppmRestApi::Config.from_json file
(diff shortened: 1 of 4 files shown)


j8r:
I'm for this change, can you detail what are the advantages for the end user? I suppose the main ones are being able to change its name and API key.

src/cli.cr
```diff
@@ -308,17 +309,17 @@ module DppmRestApi::CLI
...
                 end
-    api_key = Random::Secure.base64 24
+    group_list = split_numbers groups
+    api_key, user = DppmRestApi::Config::User.create group_list, name
     output_io.puts "API key for user named '#{name}'"
     output_io.puts api_key
     if using_stdout
+      print "ok? (Y/n) "
       raise "cancelled" if gets(chomp: true).try &.downcase.starts_with?('n')
     else
```
j8r, line 320: You can use `DPPM::CLI.confirm_prompt { raise "cancelled." }`

src/config/user.cr
```diff
@@ -48,7 +54,7 @@ struct DppmRestApi::Config::User
   end
 
   def self.deserialize(groups : String)
-    groups.split(',').map { |id| id.to_i base: 16 }
+    groups.split(',').map { |id| id.to_i base: 16 }.to_set
```
j8r: An intermediary `Array` can be avoided

src/config/user.cr
```diff
@@ -1,20 +1,24 @@
...
   API_KEY_SIZE = 63_u8
   include JSON::Serializable
-  property api_key_hash : Scrypt::Password
-  getter group_ids : Set(Int32)
-  property name : String
+  property api_key_hash, name
+  getter group_ids, id
 
-  def initialize(@api_key_hash,
+  def initialize(@api_key_hash : Scrypt::Password,
```
j8r, line 12: :ok_hand:

j8r (commented):
Please separate the formatting to a dedicated commit.
Furthermore I don't think it's necessary. Having 2 spaces results of longer lines, for me the file is a bit less readable.

src/cli.cr
dscottboggs: Nice!

src/cli.cr
```diff
@@ -36,16 +36,7 @@ module DppmRestApi::CLI
...
-              info: "The name of the users to filter by",
-            },
-            match_groups: {
-              info: "The groups that selected users must be a member of",
-            },
-            api_key: {
-              info: "The api key to select a user by. To change a user's \
-                      API key that you already know, use the rekey command",
-            },
+            user_id: "a users's UUID"
```
j8r: here is the error:
```cr
user_id: {
  info: "a user's UUID"
}
```

dscottboggs:
> Furthermore I don't think it's necessary.

I've never seen Crystal code that didn't follow two-space indentation for nested blocks.

src/config/user.cr
dscottboggs: ... but apparently shouldn't be. :thinking: 

https://github.com/DFabric/dppm-rest-api/commit/d1e005629bd7bacaacbeaa55043987aabf7f3415

j8r: No, declare a `Set` then add each `id.to_i` to it.

j8r: `groups.split(',') { ... }`

dscottboggs: Oh, I see, I forgot that `String#split` had a block variant.

```
                         iterator  43.56k ( 22.96µs) (± 4.78%)  10.7kB/op   1.23× slower
               intermediate array  45.50k ( 21.98µs) (± 3.46%)  11.5kB/op   1.18× slower
with String#split's block variant  53.78k ( 18.60µs) (± 4.60%)  7.06kB/op        fastest
```

src/config/user.cr
```diff
@@ -48,7 +55,11 @@ struct DppmRestApi::Config::User
   end
 
   def self.deserialize(groups : String)
-    groups.split(',').map { |id| id.to_i base: 16 }
+    Set(Int32).new.tap do |set|
```
j8r: I don't see the need to make a tap block here, just do `groups_set = Set(Int32).new` and an indent is saved

shard.yml
```diff
@@ -17,11 +17,6 @@ dependencies:
   dppm:
     github: DFabric/dppm
 
-development_dependencies:
-  ameba:
-    github: veelenga/ameba
-    version: ~> 0.10.0
-
```
j8r: You delete ameba?

dscottboggs: Whoops!

benchmark/benchmark_iterator_vs_intermediate_array.cr
```diff
@@ -0,0 +1,32 @@
...
+end
+
+###
+#  RESULTS
+#
+#                          iterator  43.56k ( 22.96µs) (± 4.78%)  10.7kB/op   1.23× slower
+#                intermediate array  45.50k ( 21.98µs) (± 3.46%)  11.5kB/op   1.18× slower
+# with String#split's block variant  53.78k ( 18.60µs) (± 4.60%)  7.06kB/op        fastest
+#
+###
```
j8r: To remove

shard.yml
```diff
@@ -5,23 +5,30 @@ authors:
...
 development_dependencies:
   ameba:
     github: veelenga/ameba
-    version: ~> 0.10.0
+    version: '>= 0.10.0'
 
 targets:
   dppm:
-    main: src/dppm_rest_api_cli.cr
+    main: src/dppm_rest_api_cli.cr
```
j8r: Missing newline

j8r (approved):

j8r (approved):

dscottboggs:
> I'm for this change, can you detail what are the advantages for the end user? I suppose the main ones are being able to change its name and API key.

I'm sorry that I forgot to respond to this. Prior to this, the only way a user was identified was by their name or API key, and only API keys were required to be unique. It was (and is) totally valid to have two users with the same name -- the name only functions to identify the user in the interface. Therefore, an administrator who wished to modify a user, without knowledge of their API key, would have to match the relevant users by a fuzzy-matching system which selects users based on their name and group membership. Again, these values problematically had no requirement of being unique, potentially leaving an administrator with no way to modify only one user via the commands as they were.

With this change, an administrator (or API user with administrative privileges) may still use these fuzzy matching algorithms, but only to display a selection of users information, including their UUID. Users must now be selected _by UUID_ in order to perform any mutation on that user's settings or data, allowing for no chances of difficult situations or accidental changes to multiple users.

PR #31, opened 2019-07-18
Add user/me tests

dscottboggs:
This PR adds a spec for the `GET /dppmrestapi/user/me` route, and some refactoring and revisions as a result of the testing.

### DIFF
--- a/src/config/user.cr
+++ b/src/config/user.cr
@@ -16,7 +16,7 @@ struct DppmRestApi::Config::User
   end
 
   def self.new(api_key_hash string : String,
-               groups : Set(Int32),
+               groups : Enumerable(Int32),
                name : String,
                id : UUID)
     new Scrypt::Password.new(string), groups.to_set, name, id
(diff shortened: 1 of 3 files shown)


dscottboggs:
Something about this and #25 impacts the performance of specs by an order of magnitude. :disappointed:

j8r:
Looking at the project's hisyory, the build time of this PR is on par with other recent ones.

src/config/user.cr
```diff
@@ -16,7 +16,7 @@ struct DppmRestApi::Config::User
   end
 
   def self.new(api_key_hash string : String,
-               groups : Set(Int32),
+               groups : Enumerable(Int32),
```
j8r, line 19: Why changing`Set` to `Enumerable`?

dscottboggs: Because `#to_set` is called on group before passing it to the next constructor, so this constructor can take *any* enumerable type and it will be fine.

dscottboggs:
I suspect it's the `Fixtures` change to a struct... https://cloud.drone.io/DFabric/dppm-rest-api/298/1/5

j8r:
Fixtures has to be a struct, else it won't be copied by value.

j8r:
Yes of course it is, but that's necessary - we haven't found any other solution in https://github.com/DFabric/dppm-rest-api/pull/28

j8r (approved):

PR #32, opened 2019-08-11
Fix manual testing script, allow access to /user/me by default

dscottboggs:
These changes are necessary to get the web interface working.

### DIFF
--- a/scripts/setup-for-manual-testing.sh
+++ b/scripts/setup-for-manual-testing.sh
@@ -32,12 +32,12 @@ echo '{"groups":[], "users": []}' > $DATA_DIR/permissions.json
 
 shards build
 bin/dppm server group add "name=${DPPM_USER}s group" id=$GROUP_ID
-bin/dppm server add_user "name=$DPPM_USER" groups=$GROUP_ID
+bin/dppm server user add "name=$DPPM_USER" groups=$GROUP_ID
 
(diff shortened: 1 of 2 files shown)


scripts/setup-for-manual-testing.sh
```diff
@@ -32,12 +32,12 @@ echo '{"groups":[], "users": []}' > $DATA_DIR/permissions.json
...
 bin/dppm server group add "name=${DPPM_USER}s group" id=$GROUP_ID
-bin/dppm server add_user "name=$DPPM_USER" groups=$GROUP_ID
+bin/dppm server user add "name=$DPPM_USER" groups=$GROUP_ID
 
 if [ $ADMIN = yes ]; then
   make_admin
 fi
 
-bin/dppm server run
+bin/dppm server run host="127.0.0.1"
```
j8r, line 41: The default host is already localhost, `[::1]`

dscottboggs: I realize this, but it still did not work.

Edit: I mean to say, the webpack dev server proxy couldn't connect to the running dppm server

j8r: That's a webpack configuration to do, I suppose an option like `--host ::1`.

j8r:
Script moved to https://github.com/DFabric/dppm-integration

PR #33, opened 2019-09-21
Feature: Group Actions

dscottboggs:
This PR provides access to the group controls from the REST endpoints.


j8r:
The first commit for updating mysql can be dropped now, I have pushed a commit updating dependencies today.

spec/actions/groups_spec.cr
```diff
@@ -1,34 +1,212 @@
...
 module DppmRestApi::Actions::Groups
+  struct AddGroupResponse
+    property data : NamedTuple(successfullyAddedGroup: Config::Group)
+    include JSON::Serializable
+  end
+
+  struct ChangedGroup
+    include JSON::Serializable
+    property from : Config::Group
+    property to : Config::Group
```
j8r: Can this be informative names?
I propose `source`, `destination` or `target`.

src/cli.cr
```diff
@@ -211,12 +211,17 @@ module DppmRestApi::CLI
     )
   end
 
+  # Require ALL variables
+  private macro require_all_args
+    required {{@def.args.splat}}
```
j8r: This is not really the best away to do this.
If the variables aren't optional, positional CLI arguments (`arguments`) can be used.
This isn't on this PR scope, I can submit one after.

j8r (commented):
That's a big good PR :o
Can you explain the second commit?

src/actions/groups.cr
```diff
@@ -1,34 +1,171 @@
...
+      level = context.params.url["access_level"]? || context.params.query["access_level"]?
+      super context, %[Given access level "#{level}" is not valid], cause
+    end
+  end
+
+  def query_params_in_body?(of context : HTTP::Server::Context) : Hash(String, Array(String))
+    if body = context.request.body
+      Hash(String, Array(String)).from_json body
+    else
+      {} of String => Array(String)
```
j8r:
```suggestion
     Hash(String, Array(String)).new
```
To be more consistent with the above types.

src/actions/groups.cr
```diff
@@ -1,34 +1,171 @@
...
-    # TODO: add a new route to the group
-
+    group_id_s = context.params.url["id"]
+    group_id = group_id_s.to_i? || raise BadRequest.new context,
+      %[Group ID #{group_id_s} was not a valid integer]
+    # find the group if it already exists in the permissions config
+    index = DppmRestApi.permissions_config
+      .groups
+      .index { |grp| grp.id == group_id }
+    raise NoSuchGroup.new context, group_id_s if index.nil?
```
j8r: `find` would avoid to have to do a second lookup.
The above logic can also be in a method, that can be reused the other `delete` endpoints too.

src/actions/groups.cr
```diff
@@ -1,34 +1,171 @@
...
+    end
+    # Override the existing permissions
+    group.permissions[path] = Config::Route.new access_level, existing_params
+    # and overwrite the current value.
+    DppmRestApi.permissions_config.groups[index] = group
+    build_json context.response do |response|
+      response.field "successfullyModifiedGroup" do
+        response.object do
+          response.field("from") { old_group.to_json response }
+          response.field("to") { group.to_json response }
```
j8r: An API response should better have declarative names.
I suggest `source` and `target`

src/actions/groups.cr
```diff
@@ -1,34 +1,171 @@
...
+    # existing values
+    existing_params.merge! query_params_in_body?(of: context) do |_path, existing, new_val|
+      existing + new_val
+    end
+    # Override the existing permissions
+    group.permissions[path] = Config::Route.new access_level, existing_params
+    # and overwrite the current value.
+    DppmRestApi.permissions_config.groups[index] = group
+    build_json context.response do |response|
+      response.field "successfullyModifiedGroup" do
```
j8r: I don't think it's needed, this is strange to have this as an object key.
If the response is 2**, we already know that the operation is successful.

dscottboggs:
>  Can you explain the second commit?

I believe that is leftover from something that no longer applies, I will drop it.

src/actions/groups.cr
dscottboggs: Hm. Fair enough.

j8r:
Right, the response data is not needed :+1:

src/actions/groups.cr
dscottboggs: I attempted the refactor to do this last night, and it didn't work. The problem is that since all of our data-types are structs, changing a found group doesn't update that group in the config, only the local copy of it. We need to find the index, get the value at the index, then reinsert the modified value back in... or switch to classes.

Thoughts?

src/cli.cr
dscottboggs: Relevant changes have been reverted.

j8r:
You can drop the first commit updating MySQL, already in master

shard.lock
```diff
@@ -62,7 +62,7 @@ shards:
 
   mysql:
     github: crystal-lang/crystal-mysql
-    version: 0.8.0
+    version: 0.9.0
```
j8r: You can drop the first commit updating MySQL, already in master

src/actions/groups.cr
j8r: I can't be a struct issue, nothing is passed around, and most importantly Array is always a reference.
Here the deal is because it's an array, that's why an index is needed.

dscottboggs: When you do...
```crystal
group = DppmRestApi.permissions_config
      .groups
      .find { |grp| grp.id == group_id } || raise "nil guard"
group.change_something
```
...then...
```crystal
DppmRestApi.permissions_config
      .groups
      .find { |grp| grp.id == group_id }.should_not eq group
```
...whereas if group were a class, they would still be equivalent. Does that make sense? Or am I mistaken?

src/actions/groups.cr
```diff
@@ -1,34 +1,150 @@
...
+  #  - access to "value2" and "value3" on "param4"
+  relative_delete "/:id/param/:path" do |context|
     raise Unauthorized.new context unless Actions.has_access? context, Access::Update
-    # TODO: add a new param to the group
-
+    path = context.params.url["path"]
+    group_id_s = context.params.url["id"]
+    group_id = group_id_s.to_i? || raise BadRequest.new context,
+      %[Group ID "#{group_id_s}" was not an integer!]
+    group_idx = DppmRestApi.permissions_config.groups.index { |grp| grp.id == group_id }
```
j8r: `reject!` can be used, then raise if the array is still the same.
This will also ensure that no other remaining group still have the ID.

dscottboggs: This route isn't deleting the group, it's deleting access to this path for this group. `#reject!` wouldn't work... I could see `#map` like
```crystal
group_id_already_found = false
DppmRestApi.permissions_config.groups.map do |group|
  next group if group.id != group_id
  raise InternalServerError.new context, "duplicate group with id " + group_Id_s if group_id_already_found
  group_id_already_found = true
  group.remove_access_to_path context.params.path["path"]
  group
end
DppmRestApi.permissions_config.sync_to_disk
```
... or using `#update` like `DppmRestApi.permissions_config.groups.update group_idx do |group|...`

j8r: Right, sorry

src/actions/groups.cr
```diff
@@ -1,34 +1,150 @@
...
   relative_delete "/:id" do |context|
     raise Unauthorized.new context unless Actions.has_access? context, Access::Update
-    # TODO: remove the given group
-
+    group_id = (group_id_s = context.params.url["id"]).to_i?
+    raise BadRequest.new context,
+      %[Group ID "#{group_id_s}" was not an integer] if group_id.nil?
+    to_be_deleted = DppmRestApi.permissions_config
+      .groups
+      .index { |group| group.id == group_id }
```
j8r: But here you can use `reject!`

src/actions/groups.cr
j8r: `Group#permissions` is a Hash, so a reference.
Try to use this, this will return a class instead of a copied struct:
```cr
DppmRestApi.permissions_config
      .groups
      .find { |grp| grp.id == group_id }
      .try &.permissions
```

j8r: `Group#permissions` is a Hash, so a reference.
Try to use this, this will return a class instead of a copied struct:
```cr
DppmRestApi.permissions_config
      .groups
      .find { |grp| grp.id == group_id }
      .try &.permissions
```

j8r: And what bout changing the name? The ID of a group should be immutable, but the name not.
Or, we can make `Group` totally immutable (no property), and we have to pass through `DppmRestApi::Config` to create one, that will check if the ID and the name already exist.

src/actions/groups.cr
```diff
@@ -101,50 +105,72 @@ module DppmRestApi::Actions::Groups
...
-    to_be_deleted = DppmRestApi.permissions_config
+    group_already_found = false
+    DppmRestApi.permissions_config
       .groups
-      .index { |group| group.id == group_id }
-    raise NoSuchGroup.new context, group_id_s if to_be_deleted.nil?
-    DppmRestApi.permissions_config.groups.delete_at to_be_deleted
+      .reject! do |group|
+        if group.id == group_id
+          raise DuplicateGroup.new context, group_id_s if group_already_found
```
j8r: I don't think it's good.
What does it mean for the user if (s)he gets this error, (s)he has to execute the request again?
This check of duplicated ID should be done at the group creation, only.

src/actions/groups.cr
```diff
@@ -1,34 +1,176 @@
...
+  #
+  # Not specifying a list of values...
+  # ```
+  # ?some-param&some-other-param
+  # ```
+  # ...will delete that parameter from the group.
+  #
+  # So, in summary the request...
+  # ```
+  # /123/param/%2Fsome%2Fpath?some-param=value&param2&param3&param4=value2&param4=value3
```
j8r, line 96: `%2F`? You mean `/`?

src/actions/groups.cr
```diff
@@ -1,34 +1,176 @@
...
+    group_already_found = false
+    DppmRestApi.permissions_config.groups.map! do |group|
+      next group if group.id != group_id
+      raise DuplicateGroup.new context, group_id_s if group_already_found
+      group_already_found = true
+      group.permissions.delete(
+        context.params.url["path"]
+      ) || raise NotFound.new context, "\
+                              no permissions found at the specified path \
+                              (possibly already deleted?)"
```
j8r: The phrase can start with an uppercase letter

src/actions/groups.cr
```diff
@@ -1,34 +1,176 @@
...
-    # TODO: add a new param to the group
-
+    group_id_s = context.params.url["id"]
+    group_id = group_id_s.to_i? || raise BadRequest.new context,
+      %[Group ID #{group_id_s} was not a valid integer]
+    group_already_found = false
+    path = URI.unescape context.params.url["path"]
+    access_level = Access.parse?(context.params.url["access_level"])
+    raise InvalidAccessLevel.new context if access_level.nil?
+    DppmRestApi.permissions_config.groups.map! do |group|
```
j8r, line 56: Why recreating an Array?
We just need to get the permissions Hash of a group, then modifying this reference.

src/actions/groups.cr
```diff
@@ -1,34 +1,176 @@
...
+  #  - access to "value2" and "value3" on "param4"
+  relative_delete "/:id/param/:path" do |context|
     raise Unauthorized.new context unless Actions.has_access? context, Access::Update
-    # TODO: remove the given route from the group
+    path = context.params.url["path"]
+    group_id_s = context.params.url["id"]
+    group_id = group_id_s.to_i? || raise BadRequest.new context,
+      %[Group ID "#{group_id_s}" was not an integer!]
+    group_already_found = false
+    DppmRestApi.permissions_config.groups.map do |group|
```
j8r, line 110: Same here,  `map` not needed

j8r (approved):

j8r:
Thanks @dscottboggs !

PR #34, opened 2019-09-22
Feature: App actions

dscottboggs:


src/actions/app.cr
```diff
@@ -106,12 +106,22 @@ module DppmRestApi::Actions::App
   # return the base application package
   relative_get "/:app_name/app" do |context|
     raise Unauthorized.new context unless Actions.has_access? context, Access::Read
-    # TODO: return the base application package
+    app_name = context.params.url["app_name"]
+    build_json context.response do |builder|
+      builder.field(app_name) do
+        prefix.new_app(app_name).to_json builder
```
dscottboggs: This doesn't make any sense, and the whole route should be deleted, right? Because all config data lies under `pkg`, right?

j8r: That's right for the `pkg` information, but how do we know on what package the application is based?

j8r: By the way the endpoint name is confusing, because on the filesystem `myapp/app` is a symlink to `pkg/app`.
I suggest to name it `pkg`, in reference of the `pkg` endpoint.


The actual `pkg` endpoints can be renamed to `pkg_info`.
I will make a breaking change in this sense to rename the `info` field in `pkg.con`, to avoid having `pkg_info/info` in the API.
Or do you think `pkg_file` will be fine in the API?

src/actions/app.cr
```diff
@@ -49,86 +50,174 @@ module DppmRestApi::Actions::App
...
+        false
+      else
+        # Anything but false and no get assumed to be a truthy value.
+        true
+      end
+    else
+      false
+    end
+  end
+  {% end %}
```
j8r: Please no macro,  use a method instead like `cast_bool_param(name, context)`, something like this.

src/actions/app.cr
```diff
@@ -49,86 +50,174 @@ module DppmRestApi::Actions::App
...
   relative_get "/:app_name/logs" do |context|
     raise Unauthorized.new context unless Actions.has_access? context, Access::Read
-    # TODO: upgrade to websocket or output logs to date
+    app_name = context.params.url["app_name"]
+    app = Actions.prefix.new_app app_name
+    stream_type = context.params.query["stream_type"]? || [] of String
+    stream_type = if stream_type.empty?
+                    ["errors", "output"]
+                  else
+                    ["errors", "output"] & stream_type
```
j8r: They aren't the only streams possible, and we shouldn't know about the .log extension (I push a fix upstream)

dscottboggs: Are all possible streams knowable or should we simply not filter that?

j8r: This was `App#each_log_file`, now `App#each_log_stream` in master

src/actions/app.cr
```diff
@@ -49,86 +50,177 @@ module DppmRestApi::Actions::App
...
   # if the `"stream"` query parameter is set, attempt to upgrade to a websocket
   # and stream the results. Otherwise return a JSON-formatted output of the
   # current log data.
   relative_get "/:app_name/logs" do |context|
     raise Unauthorized.new context unless Actions.has_access? context, Access::Read
-    # TODO: upgrade to websocket or output logs to date
+    app_name = context.params.url["app_name"]
+    app = Actions.prefix.new_app app_name
+    valid_streams = valid_streams_for app
+    stream_type = (context.params.query["stream_type"]? || valid_streams)
```
j8r: `stream_name` is a more suitable name

src/actions/app.cr
```diff
@@ -1,4 +1,5 @@
 require "dppm"
+# require "../config/route"
```
j8r, line 2: ?

src/actions/app.cr
dscottboggs: This already exists -- the code was simply outdated. Thank you for pointing it out.

src/actions/app.cr
dscottboggs: whoops

src/actions/app.cr
```diff
@@ -49,86 +50,139 @@ module DppmRestApi::Actions::App
...
+        app.get_logs(
+          stream_type.first,
+          follow: true,
+          lines: context.params.query["lines"]?.try &.to_i
+        ) { |line| socket.puts line }
+      end
+    else
+      JSON.build context.response do |builder|
+        builder.object do
+          builder.field "logs" do
```
j8r: Adding a field `logs` isn't needed, we already know we will get logs because of the endpoint.

j8r (commented):
What about the endpoint to get the root package and the application information?

j8r (approved):

PR #36, opened 2020-07-09
Update dependencies; Update code for compatibility with Crystal 0.35

dscottboggs:
Kemal has updated to fix a breaking change on master branch, but is not releasing a new version yet. We temporarily need to pin the master branch until they release again.

Also, `Errno` is no longer a valid stdlib type.


j8r (approved):
Thanks!

